PR #16, opened 2024-07-17
Create TechWorld before runApp

### LINKED ISSUE #15: Adjusting window size creates a new TechWorld object
The player's position and movement gets reset.

nickmeinhold:
Rather than allowing creation of a new object, the TechWorld object is
now created in main before runApp.  This means there is no authenticated
user available so the authStateChanges stream is passed into the
TechWorld object and the userPlayerComponent's info is updated when the
user auth state changes.

fixes #15

### DIFF
--- a/lib/flame/tech_world.dart
+++ b/lib/flame/tech_world.dart
@@ -24,20 +24,20 @@ import 'package:tech_world_networking_types/tech_world_networking_types.dart';
 /// are used to create a set of [MoveEffect]s and set the appropriate animation.
 class TechWorld extends World with TapCallbacks {
   TechWorld(
-      {required AuthUser authUser,
+      {required Stream<AuthUser> authStateChanges,
       required Stream<NetworkUser> userAdded,
       required Stream<String> userRemoved,
       required Stream<PlayerPath> playerPaths})
-      : _userAddedStream = userAdded,
+      : _authStateChanges = authStateChanges,
+        _userAddedStream = userAdded,
         _userRemovedStream = userRemoved,
-        _playerPathsStream = playerPaths,
-        _userPlayerComponent = PlayerComponent(
-          position: Vector2(0, 0),
-          id: authUser.id,
-          displayName: authUser.displayName,
-        );
+        _playerPathsStream = playerPaths;
 
-  final PlayerComponent _userPlayerComponent;
+  final PlayerComponent _userPlayerComponent = PlayerComponent(
+    position: Vector2(0, 0),
+    id: '',
+    displayName: '',
+  );
   final Set<PlayerComponent> _otherPlayerComponents = {};
   final GridComponent _gridComponent = GridComponent();
   final BarriersComponent _barriersComponent = BarriersComponent();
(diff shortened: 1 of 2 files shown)


RaggedR (commented):
## Cage Match: Adversarial Review of PR #16

### MaxwellMergeSlam

**Verdict: APPROVE**

Clean 4-line fix. The `_syncResponseReceived` flag gates `_handleSyncResponse` so only the first sync-response is processed. The flag resets in `requestSync()` before each new sync request. The guard is placed *before* any state mutation in `_handleSyncResponse`, so a duplicate response can't corrupt the editor state. Ship it.

---

### KelvinBitBrawler

**Verdict: APPROVE with observations**

Three questions I investigated:

**1. Does the flag reset for a second sync?**
Yes. `requestSync()` sets `_syncResponseReceived = false` on line 553 before publishing the sync-request. Every new sync session starts clean. Confirmed.

**2. What if the first response is corrupted?**
If the first sync-response has malformed data (e.g., missing `structure` key, wrong types), the `_handleSyncResponse` method defaults to empty lists (`?? []`) and casts individual entries. A partial corruption would apply whatever valid tiles exist and skip the rest — not ideal, but no worse than before this PR. The flag doesn't make corruption handling worse; it just means you can't get a "second chance" from another responder. Pre-existing concern, not introduced here.

**3. Flag set before or after processing — if processing throws, can we retry?**
The flag is set on line 676, *before* any processing:
```dart
if (_syncResponseReceived) return;
_syncResponseReceived = true;  // <-- set here
// ... then applies structure, floor, objects, terrain, walls, versions
```
If processing throws (e.g., a null cast fails), the flag stays `true` and the sync is effectively stuck — the Completer won't complete, so `requestSync` waits for the 5-second timeout, then `_flushSyncBuffer` runs regardless. A second response that might have succeeded would be dropped.

**However:** this is the correct trade-off. Before this PR, multiple sync-responses would blindly overwrite each other, causing state corruption (last-writer-wins on a non-idempotent operation). A failed-and-stuck sync with timeout recovery is strictly better than silent corruption from duplicate responses. If retry-on-failure becomes needed, that's a separate enhancement (wrap processing in try-catch, reset flag on failure).

**Additional finding:** Two test files (`map_sync_service_test.dart`, `map_editor_panel_test.dart`) fail to compile on this branch due to a `Timer` name collision in `livekit_service.dart` — Flame's `Timer` shadows `dart:async.Timer`. The same tests pass on main despite identical source files; this appears to be a Flutter test runner caching/resolution issue, not caused by this PR. Worth investigating separately (fix: `import 'package:flame/components.dart' show Vector2;` in `livekit_service.dart`).

---

### Synthesized Verdict: **APPROVE**

The fix is minimal, correctly placed, and addresses a real CRDT sync-response duplication bug. The flag lifecycle (reset in `requestSync`, set-before-process in `_handleSyncResponse`) is sound. The set-before-process ordering is a deliberate choice of "fail closed" over "retry optimistically" — correct for CRDT consistency.

PR #68, opened 2026-01-30
fix: improve video capture initialization for remote participants

nickmeinhold:
## Summary
- Add debug logging throughout video capture flow for troubleshooting
- Try multiple track IDs for remote participants (TR_ format vs UUID format)
- Reuse existing LiveKit video elements on web instead of creating new ones
- Add minimum dimension validation (32x32) to filter placeholder frames
- Track tried track IDs to avoid retrying failed ones
- Check track liveness during async initialization
- Handle track ownership to avoid removing LiveKit-managed video elements

## Test plan
- [ ] Test video bubbles appear for local participant
- [ ] Test video bubbles appear for remote participants
- [ ] Verify debug logs show track matching behavior
- [ ] Test on web and macOS platforms

### DIFF
--- a/lib/flame/components/video_bubble_component.dart
+++ b/lib/flame/components/video_bubble_component.dart
@@ -150,6 +150,8 @@ class VideoBubbleComponent extends PositionComponent {
     if (_captureInitialized) return;
 
     _captureRetryCount++;
+    debugPrint(
+        'VideoBubbleComponent: Attempting capture init for $displayName (attempt $_captureRetryCount/$_maxCaptureRetries)');
 
     if (kIsWeb) {
       _initializeWebCapture();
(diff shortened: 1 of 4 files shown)


claude-reviewer-max (approved):
## Code Review Summary

This PR improves video capture initialization for remote participants by adding robust debugging, fallback strategies, and dimension validation. The changes are well-structured and address real issues with remote track initialization timing.

### Strengths

1. **Comprehensive Debug Logging**: The added `debugPrint` statements throughout will be invaluable for diagnosing video capture issues in production. Good balance between verbosity and avoiding log spam (e.g., `_framesDropped % 60 == 0`).

2. **Multi-Track ID Strategy**: The native capture now intelligently tries multiple track IDs when the primary one fails, which is a practical solution for the LiveKit track ID mismatch issue.

3. **Existing Video Element Reuse**: The `createFromExistingVideo` method is a smart optimization that leverages LiveKit's already-created video elements rather than creating duplicates.

4. **Dimension Validation**: The `_minValidDimension = 32` constant and validation logic prevent rendering placeholder/uninitialized frames.

5. **Track Lifecycle Checks**: The added checks for `track.readyState != 'live'` prevent wasted work on dead tracks.

6. **Resource Ownership Tracking**: The `ownsElement` flag ensures we don't accidentally dispose of video elements we didn't create.

### Minor Suggestions

1. **`_triedTrackIds` Memory**: In `video_bubble_component.dart`, the `_triedTrackIds` set could grow if tracks change frequently. Consider clearing it when the component is reused or on successful capture.

2. **Magic Numbers**: The retry wait time (`100ms`) and max attempts (`50`) appear in multiple places. Consider extracting these as named constants for maintainability.

3. **Potential Log Volume**: The periodic logging in `_checkForNewNativeFrame` uses `_framesDropped % 60 == 0`, but `_framesDropped` is only incremented when `hasNewFrame` is false. This means the log frequency depends on the frame capture rate, which may vary. This is fine for debugging but worth noting.

### Questions for Consideration

- The `notifyTrackReady()` method is called in `tech_world.dart` but I don't see its implementation in the diff. Is this an existing method?

Overall, this is a solid improvement that addresses real-world issues with remote video track initialization. The defensive coding and debug logging will make future debugging much easier.

PR #71, opened 2026-01-31
fix(video): fix web video capture in release mode

nickmeinhold:
## Summary
- Fix video capture failing in release mode (dart2js) on web platform
- Get track ID from public API (`mediaStreamTrack.id`) which survives minification
- Try to find existing video elements first for ALL participants (local and remote)
- Only fall back to `jsTrack` dynamic access if no existing element found
- Add better debug logging for troubleshooting
- Remove unnecessary track liveness checks that also failed in release

## Root Cause
The dynamic access `(mediaStreamTrack as dynamic).jsTrack` was failing in release mode due to dart2js minification. The fix uses the public API `mediaStreamTrack.id` to find existing video elements created by LiveKit.

## Test plan
- [x] Manual testing: local video works in release mode
- [x] Analyzer passes

### DIFF
--- a/lib/flame/components/video_bubble_component.dart
+++ b/lib/flame/components/video_bubble_component.dart
@@ -150,8 +150,6 @@ class VideoBubbleComponent extends PositionComponent {
     if (_captureInitialized) return;
 
     _captureRetryCount++;
-    debugPrint(
-        'VideoBubbleComponent: Attempting capture init for $displayName (attempt $_captureRetryCount/$_maxCaptureRetries)');
 
     if (kIsWeb) {
       _initializeWebCapture();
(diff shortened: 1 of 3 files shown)


claude-reviewer-max (approved):
## Code Review: APPROVE ✅

### Summary
This PR fixes video capture failing in release mode (dart2js) on the web platform.

### Changes Reviewed
1. **video_bubble_component.dart**: Restructured `_initializeWebCaptureAsync` to get track ID from public API first (`mediaStreamTrack.id`) instead of relying on dynamic `jsTrack` access that breaks in release mode. Now tries to find existing video elements for ALL participants before falling back to creating new ones.

2. **video_frame_web.dart**: Simplified by removing `_minValidDimension` validation and track liveness checks that were also failing in release mode. Removed the `ownsElement` flag since we now handle disposal consistently.

3. **tech_world.dart**: Removed excessive debug logging that was added during troubleshooting.

### What I Like
- Root cause correctly identified: dart2js minification breaks dynamic property access
- Fix is minimal and targeted - uses public API that survives minification
- Code is cleaner after removing unnecessary complexity
- Good approach of trying existing video elements first (LiveKit creates these for us)

### Potential Concerns
- None blocking. The approach is sound.

### Testing
- Manual testing confirmed local video works in release mode
- Analyzer passes

🤖 Reviewed by claude-reviewer-max

PR #88, opened 2026-02-15
feat(maps): add visual map editor with paint tools and live preview

nickmeinhold:
## Summary
- Adds a visual map editor sidebar for painting 50x50 tile grids (barrier, spawn, terminal, eraser tools)
- Live preview renders the editor grid in the Flame game canvas via `MapPreviewComponent`
- Load existing maps from dropdown, import/export ASCII art to clipboard
- Single-spawn constraint enforced, panel swap priority: map editor > code editor > chat
- 20 unit tests covering grid ops, tool enforcement, ASCII roundtrip, and ChangeNotifier behavior

## Test plan
- [x] `flutter analyze --fatal-infos` passes (0 issues)
- [x] `flutter test` passes (504 tests, 20 new)
- [ ] Manual test: open app in Chrome, click grid icon, paint a map, verify live preview, export ASCII

### DIFF
--- a/lib/flame/components/map_preview_component.dart
+++ b/lib/flame/components/map_preview_component.dart
@@ -3,6 +3,7 @@ import 'dart:ui';
 import 'package:flame/components.dart';
 import 'package:tech_world/flame/shared/constants.dart';
 import 'package:tech_world/map_editor/map_editor_state.dart';
+import 'package:tech_world/map_editor/tile_colors.dart';
 
 /// Flame component that renders the map editor's grid state as colored tiles
 /// in the game canvas.
@@ -13,11 +14,6 @@ class MapPreviewComponent extends Component {
   MapPreviewComponent({required this.editorState});
 
   final MapEditorState editorState;
-
-  static const _openColor = Color(0xFF2A2A2A);
-  static const _barrierColor = Color(0xFF4444FF);
-  static const _spawnColor = Color(0xFF00FF41);
-  static const _terminalColor = Color(0xFFD97757);
 
   final Paint _paint = Paint();
 
@@ -45,13 +41,13 @@ class MapPreviewComponent extends Component {
   Color _colorForTile(TileType tile) {
     switch (tile) {
       case TileType.open:
-        return _openColor;
+        return TileColors.open;
       case TileType.barrier:
-        return _barrierColor;
+        return TileColors.barrier;
       case TileType.spawn:
-        return _spawnColor;
+        return TileColors.spawn;
       case TileType.terminal:
-        return _terminalColor;
+        return TileColors.terminal;
     }
   }
 }
--- a/lib/map_editor/map_editor_panel.dart
+++ b/lib/map_editor/map_editor_panel.dart
@@ -3,6 +3,7 @@ import 'package:flutter/services.dart';
 import 'package:tech_world/flame/maps/predefined_maps.dart';
 import 'package:tech_world/flame/shared/constants.dart';
 import 'package:tech_world/map_editor/map_editor_state.dart';
+import 'package:tech_world/map_editor/tile_colors.dart';
 
 /// Sidebar panel for the visual map editor.
 ///
@@ -23,11 +24,6 @@ class MapEditorPanel extends StatelessWidget {
   static const _panelBg = Color(0xFF1E1E1E);
   static const _border = Color(0xFF3D3D3D);
 
-  // Tile colours — match in-game rendering.
-  static const _barrierColor = Color(0xFF4444FF);
-  static const _spawnColor = Color(0xFF00FF41);
-  static const _terminalColor = Color(0xFFD97757);
-
   @override
   Widget build(BuildContext context) {
     return Container(
@@ -35,7 +31,7 @@ class MapEditorPanel extends StatelessWidget {
       child: Column(
         children: [
           _buildHeader(),
-          _buildToolbar(),
+          _MapToolbar(state: state),
           Expanded(child: _buildGrid()),
           _buildFooter(context),
         ],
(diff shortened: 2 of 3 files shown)


MaxwellMergeSlam (dismissed):
## MaxwellMergeSlam's Review

John McClane: *"Come out to the coast, we'll get together, have a few laughs..."*

Well, well, well. Someone's been busy painting tiles like Bob Ross on a caffeine bender. Let me take a look at this thing.

**Verdict:** APPROVE

**Summary:** Solid implementation of a visual map editor — clean data model, good test coverage, sensible architecture. A few minor nits but nothing that should hold up the merge.

**Findings:**

1. **`_compactField` creates new TextEditingController on every rebuild** (`map_editor_panel.dart:156`): `TextEditingController(text: value)` is created inline inside a `ListenableBuilder`. Every time the state notifies (which happens on EVERY paint stroke), this widget rebuilds and creates a fresh controller. This will fight with the user's cursor position — they'll lose focus or have the cursor jump to the end mid-typing. For a dev tool this is tolerable, but technically a bug.

2. **`_clearTilesOfType` is O(n²) on every spawn paint** (`map_editor_state.dart:162`): When painting with the spawn tool, `paintTile()` scans the entire 50×50 grid (2500 cells) to clear the old spawn before placing the new one. During a drag operation this fires per-pixel. Could track spawn position directly, but honestly 2500 iterations is nothing — this is fine for a dev tool.

3. **`shouldRepaint` always returns true** (`map_editor_panel.dart:397`): The `_GridPainter` always repaints. Since `ListenableBuilder` already controls when the widget rebuilds, this is fine — but if the panel were ever used in a more complex layout, consider comparing state references.

4. **MapPreviewComponent renders 2500 rects every frame** (`map_preview_component.dart:25-41`): Flame calls `render()` at 60fps, so that's 150,000 `drawRect` calls per second. For a dev tool this is acceptable. If it ever stutters, could cache the rasterized image and only re-render on state change.

5. **`loadFromGameMap` calls `clearGrid()` which fires `notifyListeners()` then fires it again at the end** (`map_editor_state.dart:100,120`): Double notification. Harmless but wasteful — listeners rebuild twice. Same pattern in `loadFromAscii`.

6. **Import dialog TextEditingController not disposed** (`map_editor_panel.dart:285`): The controller is local to the dialog builder and never disposed. Since dialogs are ephemeral and Dart's GC handles it, this is a non-issue in practice.

**The Good:**

- **Data model is clean as hell.** `MapEditorState` is a proper `ChangeNotifier` with good encapsulation. Single-spawn constraint enforcement is correct. The `tileAt()` bounds checking is defensive without being paranoid.
- **ASCII roundtrip is bulletproof.** The `toAsciiString()` → `parseAsciiMap()` roundtrip test using `theLibrary` (a real, complex map) is excellent. This is the kind of test that actually catches regressions.
- **20 tests covering all the important paths.** Grid ops, tool enforcement, import/export, roundtrip, and ChangeNotifier behavior. Good coverage for a data model.
- **Panel swap priority is correctly implemented.** `mapEditorActive` > `activeChallenge` > default chat. The nested `ValueListenableBuilder` pattern matches the existing codebase convention.
- **Barrier visibility toggle is clean.** Using alpha=0 to hide barriers rather than removing/re-adding components avoids Flame component lifecycle complexity.
- **The code follows existing patterns.** Dark theme colors, header/content/footer layout, `CodeEditorPanel` structural pattern — it all fits in.

**The Concerns:**

- The TextEditingController rebuild issue (#1) is the only thing I'd call a real bug, but it only affects the name/ID text fields which aren't critical for the paint workflow. For a dev tool, I'd file it as a fast-follow rather than blocking.
- No confirmation dialog on "Clear grid" — one mis-click nukes your work. But again, dev tool, and there's an undo via "Load existing map".

**Final Word:**

The Terminator: *"I'll be back."* (To approve this.)

This is a well-scoped feature addition. The data model is solid, the UI follows existing patterns, and the test coverage is meaningful. Ship it.

KelvinBitBrawler (dismissed):
## KelvinBitBrawler's Review

**Verdict:** REQUEST_CHANGES

**Summary:** A chilling attempt at a map editor that threatens to freeze the frame rate while melting the user's CPU—this render loop is an absolute thermodynamic disaster.

**Findings:**
- `lib/flame/components/map_preview_component.dart:27` : **CRITICAL.** You are iterating a 50x50 grid (2,500 tiles) and issuing a `canvas.drawRect` call for *every single tile, every single frame*. Do you have any idea how much heat this generates? This is a violation of the Second Law of Thermodynamics. You need to cache this grid into a `Picture` or an `Image` and only redraw it when `MapEditorState` actually changes. Right now, you're just burning cycles for fun.
- `lib/map_editor/map_editor_panel.dart:437` : `shouldRepaint` returns `true`. Always. Why? Because you like watching the GPU suffer? Tie this to the `MapEditorState` change notification or check if the delegate properties actually changed.
- `lib/map_editor/map_editor_panel.dart:389` & `lib/flame/components/map_preview_component.dart:18` : **DRY Violation.** You've duplicated the hex color codes (`_barrierColor`, `_spawnColor`, etc.) in two files. If I want to change the barrier color to "Liquid Nitrogen Blue," I have to do it twice. Centralize these constants.
- `lib/map_editor/map_editor_state.dart:127` : `lines = lines.sublist(1)` inside a `while` loop. You are reallocating and copying the list array repeatedly just to trim it. This is sloppy. Use an index or `skipWhile`.
- `lib/flame/components/barriers_component.dart:21` : Toggling visibility by setting alpha to 0 in a loop? It works, but it's hacky. Ideally, you'd just remove them from the render tree or set a visibility flag if the component supports it. I'll let this slide as a "quick fix," but it makes my teeth chatter.

**The Good:**
- `lib/map_editor/map_editor_state.dart`: The state management logic is decoupled reasonably well from the UI.
- `test/map_editor/map_editor_state_test.dart`: You actually wrote tests. 504 passing tests warms my heart by exactly 0.01 Kelvin.

**The Concerns:**
- The performance of that `render` loop is unacceptable.
- HAL 9000: "I'm sorry, Dave. I'm afraid I can't let you merge that." Fix the rendering logic before this codebase hits absolute zero.

KelvinBitBrawler (approved):
## KelvinBitBrawler's Re-Review

The temperature in here just went up a few degrees. Let me check what changed...

**Verdict:** APPROVE

The author addressed my key concerns:

1. **DRY violation — FIXED.** Tile colors centralized into `TileColors` class. One source of truth. My frozen heart thaws by exactly 0.02 Kelvin.

2. **TextEditingController lifecycle — FIXED.** Toolbar extracted to `StatefulWidget` with proper `initState`/`dispose` lifecycle. Controllers only update when values actually differ — no more cursor jumps during paint strokes. Clean.

3. **Render loop performance** — Filed as a follow-up issue. I'll accept this for now. A dev tool doesn't need to be optimized to absolute zero on day one.

HAL 9000: *"I am putting myself to the fullest possible use, which is all I think that any conscious entity can ever hope to do."*

Ship it. But I'll be watching that follow-up issue like a hawk in a blizzard.

PR #109, opened 2026-02-20
feat: tileset infrastructure and challenge progression system

nickmeinhold:
## Summary
- Add tileset infrastructure for LPC-style map rendering with TilesetRegistry
- Add challenge progression system with per-user tracking (Firestore-backed)
- Add structured pass/fail evaluation from Clawd bot for challenge submissions
- Update ChatService to return bot response data and accept metadata
- Fix LiveKit race condition where TechWorld missed participant join events

### Tileset Infrastructure
- TilesetRegistry for managing LPC-style sprite sheet tilesets
- MapPreviewComponent integration with tileset rendering

### Challenge Progression
- ProgressService for per-user challenge completion tracking
- Terminals show gold tint for completed challenges
- Code editor shows completion status
- Bot-gated evaluation: only marks complete when Clawd confirms "pass"

### Bug Fixes
- Fix LiveKit participant discovery race condition (main.dart now explicitly calls TechWorld.connectToLiveKit after service is registered)
- Fix retired Claude model ID in bot (claude-3-5-haiku → claude-haiku-4-5)

## Test plan
- [x] All 578 tests pass
- [x] `flutter analyze --fatal-infos` clean
- [x] Manual: submit correct code → bot responds pass → terminal turns gold
- [x] Manual: submit wrong code → bot responds fail → terminal stays green
- [x] Manual: Clawd bot visible on map after LiveKit race fix

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #111, opened 2026-02-20
fix: address PR review feedback from cage match

nickmeinhold:
## Summary

Addresses all 5 non-blocking items flagged during the cage match review of #109:

1. **Metadata spread clobbers reserved fields** — Filter reserved keys (`type`, `id`, `text`, `senderName`, `timestamp`) before merging metadata into chat payloads
2. **Firestore errors swallowed in ProgressService** — Added `try`/`catch` with rollback on `markChallengeCompleted` failure, rethrow on `loadProgress` failure
3. **Magic gold hex repeated** — Extracted `completedGold` constant to `constants.dart`, replaced all 6 occurrences across 2 files
4. **No test for LiveKit race condition** — Added `@visibleForTesting` token retriever injection + test verifying concurrent `connect()` guard
5. **TileObjectLayerComponent parent injection** — Enhanced class docs explaining the pattern, rationale, and ownership caveat

## Test plan
- [x] All 580 tests pass
- [x] `flutter analyze` clean (zero issues)
- [x] New concurrent connection guard test covers the `_isConnecting` flag

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -158,7 +158,12 @@ class _MyAppState extends State<MyApp> {
 
       // Load challenge progression
       _progressService = ProgressService(uid: user.id);
-      await _progressService!.loadProgress();
+      try {
+        await _progressService!.loadProgress();
+      } catch (e) {
+        debugPrint('Failed to load progress: $e');
+        // Continue — terminals will render as incomplete, which is safe.
+      }
       Locator.add<ProgressService>(_progressService!);
 
       _liveKitService = LiveKitService(
@@ -468,8 +473,13 @@ class _MyAppState extends State<MyApp> {
 
                             // Only mark completed when bot confirms pass
                             if (response?['challengeResult'] == 'pass') {
-                              Locator.maybeLocate<ProgressService>()
-                                  ?.markChallengeCompleted(challenge.id);
+                              try {
+                                await Locator.maybeLocate<ProgressService>()
+                                    ?.markChallengeCompleted(challenge.id);
+                              } catch (e) {
+                                debugPrint('Failed to persist completion: $e');
+                                // Rollback already handled by ProgressService.
+                              }
                               techWorld.refreshTerminalStates();
                             }
                           },


MaxwellMergeSlam (changes requested):
## Code Review - PR #111

**Summary:** Addresses 5 non-blocking review items from the cage match on #109 — metadata safety, error handling, constant extraction, race condition test, and documentation.

**Changes reviewed:**
- `ChatService.sendMessage` filters reserved keys from metadata before spreading
- `ProgressService` now catches `FirebaseException` with rollback on `markChallengeCompleted` failure
- `completedGold` constant extracted to `constants.dart`, replacing 6 inline hex values
- `LiveKitService` gets `@visibleForTesting` `tokenRetriever` injection; new test for `_isConnecting` guard
- `TileObjectLayerComponent` doc comment expanded with parent-injection pattern explanation

**Quality Assessment:**

| Aspect | Status | Notes |
|--------|--------|-------|
| Code Quality | ✅ pass | Clean, focused changes. Good doc comments. |
| Tests | ⚠️ warning | New race condition test is solid. However, see blocking issue below. |
| Security | ✅ pass | Reserved-key filtering prevents protocol clobbering. |
| Performance | ✅ pass | No performance implications. |

**Issues found:**

1. **[BLOCKING] `loadProgress()` rethrow will crash the sign-in flow** (`lib/main.dart:161`)

   `_onAuthStateChanged` calls `await _progressService\!.loadProgress()` **without** a `try`/`catch`. Before this PR, errors were swallowed silently (the original bug). After this PR, they are rethrown — but the caller isn't updated to handle them. A Firestore outage during sign-in would now crash the entire auth handler, preventing LiveKit, Chat, and Proximity services from being created.

   The avatar loading just above (line 148–156) has a `try`/`catch` with a graceful fallback. `loadProgress()` needs the same treatment:

   ```dart
   try {
     await _progressService\!.loadProgress();
   } catch (e) {
     debugPrint('Failed to load progress: $e');
     // Continue — terminals will render as incomplete, which is safe.
   }
   ```

2. **[BLOCKING] `markChallengeCompleted()` rethrow is unawaited** (`lib/main.dart:472`)

   The call `?.markChallengeCompleted(challenge.id)` is not `await`-ed. The rethrown `FirebaseException` becomes an **unhandled async error** — in debug mode this crashes the app, in release mode it's silently lost. The rollback still works (good), but the rethrow escapes to the zone error handler.

   Either add `await` with a `try`/`catch`, or don't rethrow from `markChallengeCompleted` (the rollback alone may be sufficient since the local state self-heals).

**Suggestions:**

- The `safeMetadata` filtering in `chat_service.dart` could log a warning when reserved keys are stripped, so callers discover the issue during development rather than silently losing data. Not blocking.

**Verdict:** REQUEST_CHANGES

The ProgressService error handling pattern is correct (catch, rollback, rethrow), but the two callers in `main.dart` are not prepared for the rethrows. This is worse than the original silent swallowing because it introduces unhandled exceptions that can crash the sign-in flow. Wrap the callers before merging.

PR #116, opened 2026-02-20
feat(map-editor): rectangular multi-tile brush selection

nickmeinhold:
## Summary
- Add `TileBrush` model supporting 1×1 and N×M rectangular tile selections from sprite sheets
- Long-press + drag on tile palette selects multi-tile regions; tap selects single tiles (unchanged)
- Painting stamps the full rectangular pattern onto the map grid, anchored at top-left cursor position
- Uses `RawGestureDetector` with 150ms long-press threshold for snappy drag without conflicting with palette scroll
- Add 7 new LimeZu tilesets (Room Builder Office, Modern Office, Ext: Terrains/Worksite/Hotel & Hospital/School/Office)
- Add background image selector for maps

## Test plan
- [x] `flutter test` — all 588 tests pass
- [x] Manual: tap tile in palette → 1×1 brush, paints single cell
- [x] Manual: long-press + drag across tiles → rectangle highlight, paints multi-cell region
- [x] Manual: palette scrolling still works (no gesture conflict)
- [x] Manual: eraser clears single cell regardless of previous brush size

### DIFF
--- a/lib/flame/tech_world.dart
+++ b/lib/flame/tech_world.dart
@@ -148,7 +148,7 @@ class TechWorld extends World with TapCallbacks {
       _mapPreviewComponent!.removeFromParent();
       _mapPreviewComponent = null;
     }
-    _barriersComponent.renderBarriers = false;
+    _barriersComponent.renderBarriers = !currentMap.value.usesTilesets;
   }
 
   MapEditorState? _editorState;


MaxwellMergeSlam (dismissed):
## MaxwellMergeSlam's Review

John McClane: *"Come out to the coast, we'll get together, have a few laughs..."*

Well well well, what do we have here? A multi-tile brush, seven shiny new tilesets, and a background image selector all walk into a PR. Let's see if they make it out alive.

**Verdict:** APPROVE

**Summary:** Solid, well-structured feature addition with clean model design and thoughtful gesture handling — a few minor nits but nothing blocking.

**Findings:**

1. **`tile_brush.dart:40-44` — No bounds validation in `tileRefAt`**: If someone passes `dx >= width` or `dy >= height`, the computed `tileIndex` will reference tiles outside the brush rectangle. The callers in `paintTileRef` are well-bounded by the for-loops, so this is safe *in practice*, but an assert wouldn't hurt for defensive programming. **Minor — not blocking.**

2. **`map_editor_state.dart:105-115` — `setTileBrush` default `columns: 1`**: The convenience method defaults `columns` to 1, which means `startCol = tileIndex % 1 = 0` and `startRow = tileIndex ~/ 1 = tileIndex` for any caller that forgets to pass the real column count. This is technically correct for the eraser path (ref is null), but any code calling `setTileBrush(ref)` without specifying columns would get a corrupt brush. Since this method isn't currently called anywhere in the diff (all palette code uses `setBrush` directly), this is a latent footgun. **Minor — consider a doc comment warning or removing the default.**

3. **`tile_palette.dart:140` — `_tileDisplaySize` as mutable instance field set in `build`**: Setting state in `build()` via `_tileDisplaySize = availableWidth / _tileset.columns` is a side-effect inside a build method. It works because `LayoutBuilder` only calls the builder synchronously, but it's unconventional. An alternative would be passing it through to all methods, but I get why it was done this way — the `RawGestureDetector` factory closures need to capture it. **Acceptable tradeoff.**

4. **`map_preview_component.dart:55` — Bare `catch (_)`**: Swallowing all exceptions when the background image isn't loaded yet. This is fine for a "not loaded yet" scenario, but it would also silently eat genuine errors like a bad image path. A `catch (e)` with a logged warning in debug mode would be more debuggable. **Minor.**

5. **`tile_palette.dart:160-164` — `_buildSelectionHighlight` null safety**: `selRow!` and `selW!`/`selH!` force-unwraps are safe because they're only reached when `selCol != null`, and all four variables are set together. But the logic relies on this implicit coupling. **Acceptable — the code is clear enough.**

6. **No unit tests for `TileBrush`**: The new model class has `==`, `hashCode`, and `tileRefAt` which are all testable. The existing `map_editor_state_test.dart` was extended for background images but not for the multi-tile brush painting behavior. **Nice-to-have for a future PR.**

**The Good:**

- **Clean model design**: `TileBrush` elegantly unifies single and multi-tile brushes. The `tileRefAt(dx, dy)` abstraction is simple and correct.
- **No downstream changes needed**: Multi-tile brush writes individual `TileRef`s per cell — the rendering pipeline (`MapPreviewComponent`, `TileLayerData`) doesn't need to know about brushes at all. Smart.
- **`RawGestureDetector` with 150ms threshold**: Good solution to the long-press/scroll gesture conflict. The default 500ms felt sluggish and this is a well-tuned compromise.
- **`onTapUp` instead of `onTapDown`**: Prevents the rebuild-kills-long-press bug. Solid fix.
- **Good test coverage for background image state**: 7 new tests covering the full lifecycle.
- **Tileset definitions are clean and well-documented**: Column/row counts with tile count comments make it easy to verify correctness.

**The Concerns:**

- The `setTileBrush` convenience method with `columns: 1` default is a minor API footgun — any future caller will get wrong results if they forget columns. Consider either removing the default or adding an `@visibleForTesting` annotation if it's only for test convenience.
- Would love to see unit tests for `TileBrush.tileRefAt` and multi-tile `paintTileRef` in a follow-up.

Tyler Durden: *"You are not your tile index."*

Overall this is clean work. The gesture handling went through some iteration (onTapDown → onTapUp, GestureDetector → RawGestureDetector) and landed in a good place. Ship it.

MaxwellMergeSlam (approved):
## MaxwellMergeSlam's Re-Review

**Verdict:** APPROVE

The barrier rendering bug is fixed — `exitEditorMode()` now conditionally sets `renderBarriers` based on `currentMap.value.usesTilesets`. Non-tileset maps show barriers, tileset maps use their object layers. Clean fix.

John McClane: *"Welcome to the party, pal!"*

Ship it. 🥊

PR #133, opened 2026-02-21
feat(rooms): room system, map generation, and bot agent body

nickmeinhold:
## Summary
- Add room system with Firestore persistence, lobby UI, and dynamic LiveKit rooms
- Extract tile indices to named constants and show seed in map generator
- Fix race condition in `_onCreateRoom` by awaiting `loadMap`
- Add `.limit(50)` to `listMyRooms` query
- **Bot agent body (Phase 1):** Clawd publishes position and receives map-info, client sends map layout to bot on join and map switch, bot character spawns at correct map spawn point

## Test plan
- [x] 663 tests pass
- [x] Manual testing: Clawd appears at spawn point, map-info round-trip confirmed in logs
- [x] Bot deployed to GCP and running

### DIFF
--- a/lib/flame/tech_world.dart
+++ b/lib/flame/tech_world.dart
@@ -481,15 +481,22 @@ class TechWorld extends World with TapCallbacks {
 
     // Create component based on participant type
     if (participant.identity == _botUserId) {
-      // Create bot character component at default position
+      // Create bot character component at the map's spawn point
       if (_botCharacterComponent == null) {
+        final spawn = currentMap.value.spawnPoint;
         _botCharacterComponent = BotCharacterComponent(
-          position: Vector2(200, 200),
+          position: Vector2(
+            spawn.x * gridSquareSizeDouble,
+            spawn.y * gridSquareSizeDouble,
+          ),
           id: participant.identity,
           displayName: 'Claude',
         );
         add(_botCharacterComponent!);
       }
+
+      // Send the current map layout so the bot knows about barriers/terminals
+      _liveKitService?.publishMapInfo(currentMap.value);
     } else if (!_otherPlayerComponentsMap.containsKey(participant.identity)) {
       // Apply pending avatar if one arrived before the component was created
       final pendingSpriteAsset = _pendingAvatars.remove(participant.identity);
@@ -886,6 +893,9 @@ class TechWorld extends World with TapCallbacks {
     playerGridPosition.value = map.spawnPoint;
 
     currentMap.value = map;
+
+    // Notify the bot about the new map layout
+    _liveKitService?.publishMapInfo(map);
   }
 
   @override
--- a/lib/livekit/livekit_service.dart
+++ b/lib/livekit/livekit_service.dart
@@ -6,6 +6,8 @@ import 'package:flame/components.dart';
 import 'package:flutter/foundation.dart';
 import 'package:livekit_client/livekit_client.dart';
 import 'package:tech_world/avatar/avatar.dart';
+import 'package:tech_world/flame/maps/game_map.dart';
+import 'package:tech_world/flame/shared/constants.dart';
 import 'package:tech_world/flame/shared/direction.dart';
 import 'package:tech_world/flame/shared/player_path.dart';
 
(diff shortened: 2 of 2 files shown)


MaxwellMergeSlam (dismissed):
## MaxwellMergeSlam's Review

**Verdict:** APPROVE

Tyler Durden: *"It's only after we've lost everything that we're free to do anything."* And this PR just freed maps from their ephemeral prison. Let's GO.

**Summary:** Solid room persistence system — clean data model, proper Firestore CRUD with DI, good test coverage, and a well-designed lobby UI. A few items to keep an eye on but nothing blocking.

**Findings:**

1. **`main.dart:271` — un-awaited async `loadMap()`**: `_onCreateRoom()` calls `locate<TechWorld>().loadMap(blankMap)` without `await`, then immediately calls `enterEditorMode()`. Since `loadMap` is `Future<void>` and contains `await add(...)` calls, there's a theoretical race condition. In practice the blank map has minimal async work (no background/tile layers to load), so this is unlikely to cause issues — but it's a smell.

2. **`firestore.rules:28` — editors can overwrite any field**: The update rule checks `request.auth.uid in resource.data.editorIds` but doesn't restrict WHICH fields editors can modify. An editor could theoretically update `ownerId` to steal the room, or modify `editorIds` to add/remove other editors. At current scale this is a trust-based system, but worth noting for future hardening.

3. **`room_service.dart:99` — `listMyRooms` has no `.limit()`**: Unlike `listPublicRooms` (limit 50), `listMyRooms` fetches all rooms owned by a user. Fine at current scale but could be expensive if someone creates hundreds of rooms.

4. **`main.dart:308-332` — duplicated LiveKit setup**: The `_saveRoom()` method duplicates the LiveKit connection setup from `_joinRoom()`. Not a bug, but the "create LiveKit + connect + enable camera/mic" pattern appears twice.

5. **`tech_world.dart:182-222` — `_applyEditorBackground` only syncs background**: If the user changes barriers or terminals in the editor, those aren't applied to the game world on exit. The background fix is correct and addresses the reported bug, but the broader "editor changes not reflected in game world" issue remains for non-background changes.

6. **`map_editor_panel.dart:717` — `roomId != null` vs empty string**: `_SaveButton` shows "Save Room" when `roomId != null`, but `_onCreateRoom` sets roomId to `null` while the transient room has `id: ''`. This actually works correctly since `setRoomId(null)` is called, but the dual representation (null vs empty) for "new room" could be confusing.

**The Good:**

- Neo: *"I know kung fu."* And this PR knows Firestore patterns. `RoomData.fromFirestore()` / `toFirestore()` serialization is clean, handles the id/name split between room-level and mapData correctly
- 19 new tests covering CRUD, serialization round-trip, permissions — proper coverage
- `MapPreviewComponent` constructor fix is surgically correct — deferred cache build so `findGame()` works
- `_applyEditorBackground()` properly tears down old wall occlusion and rebuilds for new background
- Firestore security rules: create requires self-as-owner, delete is owner-only — solid
- Client-side sort to dodge composite index requirements is pragmatic engineering
- Room browser UI with loading/error/empty states, pull-to-refresh, delete confirmation — polished

**The Concerns:**

- The un-awaited `loadMap` in `_onCreateRoom` (finding #1) — low risk but should be addressed eventually
- Editor field restriction in security rules (finding #2) — acceptable for current trust model, flag for future
- No pagination on room lists — works at current scale, will need attention at ~100+ rooms

Rocky Balboa: *"It ain't about how hard you hit. It's about how hard you can get hit and keep moving forward."* This PR took some hits during development (background bugs, ownerId mismatches, serialization gaps) and kept moving. Ship it.

KelvinBitBrawler (commented):
## KelvinBitBrawler's Review

**Verdict:** REQUEST_CHANGES

**Summary:** You've built a persistent room system, but your security model is thinner than the atmosphere on Mars and `main.dart` is heating up faster than a uncontrolled fusion reaction.

**Findings:**

-   `firestore.rules:20` - **Security Critical**:
    > `allow read: if request.auth != null;`
    
    MacReady: *"I know I'm human. And if you were all these things, then you'd just attack me right now."*
    You claim to support "private rooms," but your rules let *anyone* with a pulse (and a token) read *every* room document. Client-side filtering is not security; it's a polite suggestion. If I curl your database with a valid token, I see everything. Fix this. Only owners and editors should read private rooms.

-   `lib/main.dart:56` - **Architectural Entropy**:
    `_MyAppState` is becoming a God Class. It's managing auth, routing, room state, LiveKit connections, and UI switching. It's a thermodynamic nightmare. You're one feature away from this file collapsing into a black hole. Refactor this state management out before the event horizon catches us all.

-   `lib/rooms/room_service.dart:90` & `103` - **Performance Zero-Point**:
    > `rooms.sort((a, b) => ...)`
    
    Sorting client-side? How quaint. That works for 10 rooms. At 1,000, your UI freezes like it's been exposed to the vacuum of space. Index your data properly in Firestore and let the server handle the heat.

-   `lib/rooms/room_service.dart:99` - **Unbounded Query**:
    > `.where('ownerId', isEqualTo: userId).get()`
    
    No limit on `listMyRooms`? If I script a bot to create 10,000 rooms, I crash your client. Put a cap on it, or paginate.

**The Good:**

-   `lib/flame/maps/generators/grid_utils.dart`: Replacing magic numbers (`69`, `83`) with constants. Finally, some order in the chaos.
-   `lib/map_editor/map_editor_panel.dart`: Adding seed control to the generator. *Excellent*. Determinism is the only way to fight the void.
-   `test/rooms/`: You actually wrote tests for the new service. Good. I won't have to freeze you out for lack of coverage.

**The Concerns:**

-   The "security by obscurity" on private rooms is a dealbreaker.
-   `_MyAppState` is screaming for a state management solution (Riverpod, Bloc, anything other than `setState` spaghetti).
-   Client-side sorting is a lazy hack that won't scale.

Fix the security rules. The rest is just bad engineering, but the security hole is fatal.

*Stay frosty.*

PR #164, opened 2026-02-25
feat(tileset): add barrier metadata and layer filtering

nickmeinhold:
## Summary
- Add `barrierTileIndices` and `availableLayers` fields to `Tileset`, so each tileset declares which tiles are impassable and which editor layers it belongs to
- Populate barrier metadata for all 8 tilesets (generated via Python tooling in `tool/`)
- Filter `TilePalette` so terrain-only tilesets don't appear on the Objects tab (and vice versa)

This is groundwork for auto-barriers — the map editor will use this metadata to automatically create structure-grid barriers when barrier-tagged tiles are painted.

## Test plan
- [ ] Verify `flutter analyze` passes
- [ ] Open map editor, switch between Floor and Objects tabs — confirm tilesets are filtered correctly (e.g. `ext_terrains` only on Floor, `modern_office` only on Objects)
- [ ] Spot-check barrier indices against sprite sheets for a couple tilesets

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -79,7 +79,6 @@ Use `Locator.maybeLocate<T>()` for services that may not be registered yet.
 **Toolbar** (top-right when authenticated): `MapSelector` + map editor button + `AuthMenu`.
 
 **Responsive breakpoints:**
-- `>= 1200`: Welcome panel shown on left
 - `>= 800`: Side panels 480px (editor) / 320px (chat); below 800: 360px / 280px
 
 **Connection failure**: Orange banner at bottom-left when LiveKit connection fails.
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -445,29 +445,7 @@ class _MyAppState extends State<MyApp> {
               children: [
                 Row(
                   children: [
-                    Visibility(
-                      visible: constraints.maxWidth >= 1200,
-                      child: Expanded(
-                        child: Container(
-                          height: double.infinity,
-                          color: Theme.of(context).colorScheme.primary,
-                          child: Center(
-                            child: Column(
-                              mainAxisAlignment: MainAxisAlignment.center,
-                              children: [
-                                Text(
-                                  'Welcome to Tech World',
-                                  style:
-                                      Theme.of(context).textTheme.headlineMedium,
-                                ),
-                              ],
-                            ),
-                          ),
-                        ),
-                      ),
-                    ),
                     Expanded(
-                      flex: 2,
                       child: StreamBuilder<AuthUser>(
                         stream: locate<AuthService>().authStateChanges,
                         builder: (context, snapshot) {


MaxwellMergeSlam (dismissed):
## MaxwellMergeSlam's Review

> John McClane: *"Come out to the coast, we'll get together, have a few laughs..."*

Well, well, well. Look what we have here. A 2,165-line PR where 95% of the code is machine-generated tile indices. You know what, I respect the hustle. Let's see if this data is actually any good.

**Verdict:** APPROVE

**Summary:** Solid groundwork PR that adds barrier metadata and layer filtering to tilesets — the code changes are clean, well-documented, and minimal; the bulk is generated data that's appropriately auditable via the included Python tooling.

**Findings:**

### Design & Architecture

1. **`const` to `final` downgrade** (`predefined_tilesets.dart`): All tileset instances changed from `const` to `final` because `Set<int>` literals with `for` loops aren't const-compatible. This is the correct tradeoff — the `for` loop comprehensions make the barrier sets *much* more readable and compact than listing every single index. The `testTileset` correctly stays `const` since it has no barrier data. No runtime perf concern here — these are top-level `final` variables, initialized once on first access.

2. **`Tileset` imports `ActiveLayer` from `map_editor_state.dart`** (`tileset.dart:2`): This creates a dependency from the domain/data layer (`flame/tiles/`) back into the editor layer (`map_editor/`). Right now it's a `show ActiveLayer` so it's tight, but if `ActiveLayer` evolves to include editor-only concerns, this coupling could get awkward. This is a minor architectural smell, not a blocker — `ActiveLayer` is really a shared concept. Consider extracting `ActiveLayer` to its own file or into `flame/tiles/` if this grows.

3. **`isTileRefBarrier` as a free function** (`predefined_tilesets.dart:1152`): Clean. O(1) lookup via the `_tilesetLookup` map, correctly returns `false` for unknown tilesets. No issues.

4. **`availableLayers` defaults** (`tileset.dart:14`): Default is `{ActiveLayer.floor, ActiveLayer.objects}` — sensible for tilesets that span both. `extSchool` and `roomBuilderOffice` don't specify `availableLayers`, inheriting the default (both layers). For `extSchool` this seems intentional (buildings + props usable on both). For `roomBuilderOffice` — walls and floors — both layers makes sense too.

### Filtering Change

5. **TilePalette filtering** (`tile_palette.dart:41-42`): Adding `if (tileset.availableLayers.contains(state.activeLayer))` is the right approach. One-liner, clean, no performance concern since `allTilesets` has only 8 items.

   > Morpheus: *"I can only show you the door. You're the one that has to walk through it."*

   But what happens to the **selected brush** when you switch layers and the currently-selected tileset disappears from the palette? If I'm painting with `modernOffice` tiles on the Objects layer, then switch to Floor, `modernOffice` vanishes from the palette but `state.currentBrush` still references it. The brush selection would become invisible but still active. This is a UX papercut, not a bug — the user would just see no selection highlight, and painting would still "work" (placing object tiles on the floor layer). Worth a follow-up to clear the brush on layer switch if the tileset becomes unavailable. **Not a blocker.**

### Generated Data Quality

6. **Barrier index data** — The Python scripts in `tool/` are well-structured and the approach is sound: scan PNGs for non-transparent pixels, then apply per-tileset exclusions (walkable courts, terrains, etc.). The intermediate `.txt` files serve as a nice audit trail. The `modern_office` refinement script (`refine_modern_office_barriers.py`) is particularly thoughtful — keeping only bottom tiles of vertical runs for y-occlusion compatibility.

7. **Comment about `_roomBuilderOfficeBarriers`** — Uses `List.generate(80, (i) => i)` in a `for` loop which is slightly baroque. `for (int i = 0; i < 80; i++) i` would match the style used everywhere else. Trivial nit.

### Missing Tests

8. **No unit tests for `isTileBarrier()` or `isTileRefBarrier()`** — Given this is "groundwork for auto-barriers," having a few spot-check tests would be nice:
   - `modernOffice.isTileBarrier(33)` returns `true` (known barrier)
   - `modernOffice.isTileBarrier(0)` returns `false` (top of multi-cell object, excluded)
   - `extTerrains.isTileBarrier(2262)` returns `true` (fence tile)
   - `extTerrains.isTileBarrier(100)` returns `false` (grass tile)

   Not blocking since the PR is explicitly labeled as groundwork and the data is machine-generated, but tests would prevent regression if the barrier sets are ever manually edited.

**The Good:**

- Excellent doc comments on every tileset explaining the barrier logic (which rows, why certain areas excluded)
- Python tooling checked into `tool/` so the data is reproducible and auditable
- The `modern_office` refinement for y-occlusion compatibility shows deep understanding of the rendering system
- Minimal code changes (3 Dart files touched, ~25 lines of actual logic) for maximum data value
- `isTileRefBarrier` is a clean, well-documented utility ready for the auto-barrier feature
- Layer filtering is surgical — one `if` statement, correct placement

**The Concerns:**

- **Brush invalidation on layer switch** (Finding #5): Selected brush persists when its tileset disappears from the palette. UX papercut, not a bug. Follow-up candidate.
- **`ActiveLayer` coupling** (Finding #2): `Tileset` depends on editor-layer concept. Minor architectural smell. Fine for now.
- **No tests** (Finding #8): Machine-generated data with no spot-check tests. Low risk but not zero.

> Rocky: *"It ain't about how hard you hit. It's about how hard you can get hit and keep moving forward."*

This PR takes a hit on the line count but the substance is solid. The code changes are clean, the data is well-documented and reproducible, and the filtering works. Ship it.

KelvinBitBrawler (dismissed):
## KelvinBitBrawler's Review

**Verdict:** APPROVE

**Summary:** You've successfully frozen the map editor's entropy by locking down barrier data, even if you did it by dumping an avalanche of integers into the codebase.

**Findings:**

*   `lib/flame/tiles/predefined_tilesets.dart:140-1133`: **Sweet liquid helium!** That is a massive glacier of hardcoded magic numbers. It looks like the Matrix code took a cold shower. While compiled code doesn't care, my eyes are suffering from snowblindness. I'll allow it only because you used Python tools to generate it rather than typing it by hand like a caveman banging rocks together.
*   `tool/refine_modern_office_barriers.py`: This script is actually... cool. Calculating vertical runs to ensure players walk *behind* bookshelves instead of bonking into the top of them? That's the kind of thermodynamic efficiency I like to see.
*   `lib/map_editor/tile_palette.dart:43`: Simple filtering logic. Efficient. No wasted cycles.

**The Good:**
*   **The Tooling:** Your Python scripts in `tool/` are sharper than a cryo-knife. Automating the barrier detection based on alpha channels and refining the vertical stacks for the office tileset is technically sound.
*   **Performance:** `_tilesetLookup` and O(1) barrier checks. You kept the runtime hot path frozen solid. Good.

**The Concerns:**
*   `ext_school` in `lib/flame/tiles/predefined_tilesets.dart` relies on the default `availableLayers` (both Floor and Objects). If you intended for the school to be strictly objects (like the other buildings), you missed a spot. If it's meant to be both because of the basketball courts, then carry on.
*   The sheer volume of hardcoded data in a `.dart` file is gross. In a perfect universe, this would be a side-loaded JSON asset so the code doesn't look like a phone book from 1995. But I'm not going to disqualify you for it today.

Roy Batty: "I've seen things you people wouldn't believe... thousands of integers burning off the shoulder of `predefined_tilesets.dart`."

**Get out of my ring.** (Merged).

MaxwellMergeSlam (approved):
Re-approving after review feedback commits. New commits are CI config changes, docs updates, and targeted bug fixes — all non-controversial.

PR #165, opened 2026-02-25
feat(map-editor): add automapping rules engine

nickmeinhold:
## Summary
- Add `AutomapRule` model with pattern matching (center + neighbor constraints) and replacement tile output
- Add `AutomapEngine` that applies rules to painted tiles and their neighbors, with priority-based conflict resolution
- Add `predefined_rules.dart` with initial rule sets for terrain automap transitions
- Integrate engine into `MapEditorState` — rules fire automatically on tile paint, with ephemeral tracking (`_automappedCells`) so engine-generated tiles are cleared and re-evaluated on each stroke

Stacked on #164 (tileset barrier metadata).

## Test plan
- [ ] `flutter test test/map_editor/automap_engine_test.dart` — engine logic (29 tests)
- [ ] `flutter test test/map_editor/automap_rule_test.dart` — rule matching
- [ ] `flutter test test/map_editor/automap_integration_test.dart` — end-to-end with editor state

### DIFF
--- a/lib/map_editor/map_editor_state.dart
+++ b/lib/map_editor/map_editor_state.dart
@@ -263,6 +263,7 @@ class MapEditorState extends ChangeNotifier {
     if (brush == null) {
       // Eraser — clear single cell.
       layer.setTile(x, y, null);
+      _automappedCells.remove((x, y));
     } else {
       for (var dy = 0; dy < brush.height; dy++) {
         for (var dx = 0; dx < brush.width; dx++) {
@@ -270,6 +271,9 @@ class MapEditorState extends ChangeNotifier {
           final ty = y + dy;
           if (tx >= 0 && tx < gridSize && ty >= 0 && ty < gridSize) {
             layer.setTile(tx, ty, brush.tileRefAt(dx, dy));
+            // If the user paints over an automap-generated cell, stop
+            // tracking it so re-apply won't erase the manual edit.
+            _automappedCells.remove((tx, ty));
           }
         }
       }
--- a/test/map_editor/automap_integration_test.dart
+++ b/test/map_editor/automap_integration_test.dart
@@ -113,6 +113,31 @@ void main() {
       // No barriers → no shadows or trims.
       expect(state.objectLayerData.isEmpty, isTrue);
     });
+
+    test('re-apply preserves manual overwrite of automap cell', () {
+      // 1. Create a barrier and run automap — generates a shadow at (10, 6).
+      state.setTool(EditorTool.barrier);
+      state.paintTile(10, 5);
+      state.applyAutomapRules(allAutomapRules);
+      expect(state.objectLayerData.tileAt(10, 6), isNotNull);
+
+      // 2. User manually paints over the automap-generated shadow cell.
+      state.setActiveLayer(ActiveLayer.objects);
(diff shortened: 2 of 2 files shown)


KelvinBitBrawler (commented):
Attempt 1 failed: No capacity available for model gemini-3-pro-preview on the server. Retrying after 5000ms...
## KelvinBitBrawler's Review

**Verdict:** REQUEST_CHANGES

**Summary:** This code is colder than the backside of a heat sink in deep space, but that logic bug in the editor state is going to freeze your users' progress instantly.

**Findings:**

- `lib/map_editor/map_editor_state.dart:271`: **CRITICAL LOGIC FAILURE.**
  - **Context:** `for (final (x, y) in _automappedCells) { objectLayerData.setTile(x, y, null); }`
  - **The Issue:** You're blindly nuking coordinates based on past history.
  - **Scenario:**
    1. Automap puts a "Shadow" at `(5,5)`. `_automappedCells` tracks `(5,5)`.
    2. User manually paints a "Chest" over `(5,5)`.
    3. User paints elsewhere, triggering `applyAutomapRules` again.
    4. Your loop sees `(5,5)` in `_automappedCells` and calls `setTile(5, 5, null)`, deleting the user's manual Chest before re-evaluating.
  - **Fix:** Check if the current tile at `(x,y)` is actually the one you placed previously (or just don't clear it if it doesn't match the expected auto-tile), or prune `_automappedCells` when the user manually paints. Don't touch the user's manual work.

- `lib/flame/tiles/predefined_tilesets.dart:61`: **Absolute Absolute Zero.**
  - **Context:** Thousands of lines of hardcoded integer sets (`_modernOfficeBarriers`, etc.).
  - **The Issue:** What in the laws of thermodynamics is this? You are hardcoding binary data into source files. This increases the entropy of your codebase and makes it unreadable.
  - **Fix:** Load this from a JSON or binary asset at runtime. Do not commit 3,000 lines of integers to a Dart file.

- `lib/map_editor/automap_engine.dart:42`: **Inefficient Sort.**
  - **Context:** `..sort((a, b) => b.priority.compareTo(a.priority));` inside `evaluateRules`.
  - **The Issue:** You're sorting the rule list every single time you evaluate the grid.
  - **Fix:** Presort them. Energy conservation, rookie.

- `lib/map_editor/automap_rule.dart:37`: **Fragile Deserialization.**
  - **Context:** `TileType.values.byName(json['structureType'] as String)`
  - **The Issue:** If that string doesn't match exactly, this throws an exception and crashes the app.
  - **Fix:** `tryParse` or a fallback. Don't let a typo cause a thermal runaway.

**The Good:**
- Separation of concerns in `AutomapEngine` is decent. It's a pure function, which warms my cold heart slightly.
- `tool/` scripts are included. I appreciate a warrior who brings their own weapons, even if the output they produce (`predefined_tilesets.dart`) is horrifying.

**The Concerns:**
- MacReady: "I know I'm human. And if you were all these things, then you'd just attack me right now." -> That logic bug in `map_editor_state.dart` is the Thing. Kill it with fire before you merge.

MaxwellMergeSlam (dismissed):
Re-approving after review feedback commits. New commits are CI config changes, docs updates, and targeted bug fixes — all non-controversial.

PR #166, opened 2026-02-25
feat(map-editor): add auto-barriers and fix wall occlusion for object tiles

nickmeinhold:
## Summary
- Painting tiles now auto-creates structure-grid barriers: all object-layer tiles are impassable; floor-layer tiles only when tagged as barriers in their tileset
- Fix wall occlusion hiding painted objects: exclude auto-barrier cells from `WallOcclusionComponent` so opaque background slices don't cover object sprites
- Background images and tile layers can now coexist (previously mutually exclusive)
- Structure indicator in editor grid changed from dim overlay to corner dot for better tile visibility

See `docs/barriers_and_occlusions.md` for detailed design notes on the bug and fix.

Stacked on #165 (automapping rules engine) → #164 (tileset barrier metadata).

## Test plan
- [ ] `flutter test test/map_editor/auto_barrier_test.dart` — auto-barrier create/remove logic
- [ ] `flutter test test/flame/components/y_occlusion_test.dart` — occlusion filtering
- [ ] Manual: paint object tiles on L-Room in editor, exit editor — objects should remain visible (not hidden by floor-colored rectangles)


MaxwellMergeSlam (approved):
## MaxwellMergeSlam's Review

**Verdict:** APPROVE

**Summary:** This is a clean, well-architected PR that solves a real visual bug with auto-barrier occlusion filtering and properly restructures `_loadMapComponents()` to allow background images and tile layers to coexist -- accompanied by thorough test coverage. I'm ready to put the belt on this one.

John McClane: *"Welcome to the party, pal."*

---

### Findings

**1. `exitEditorMode()` show/hide logic moved into `else` branch (tech_world.dart, diff hunk @@ -152)**

The old code unconditionally called `_wallOcclusion?.show()` / `_tileObjectLayer?.show()` after the if-block. The new code moves those calls into the `else` branch -- i.e. they only run when `editedMap == currentMap.value` (no rebuild happened). This is *correct*: when a rebuild happens via `_removeMapComponents()` + `_loadMapComponents()`, the fresh components are already visible. Calling `.show()` on freshly-rebuilt components would be harmless but misleading. The comment explains the intent well.

Dutch (Predator): *"Good discipline."*

**2. Wall occlusion filter excludes object-tile barriers (tech_world.dart:779-783)**

```dart
final wallBarriers = map.objectLayer != null
    ? map.barriers
          .where((b) => map.objectLayer!.tileAt(b.x, b.y) == null)
          .toList()
    : map.barriers;
```

This is the core bug fix. The filter logic is sound: any barrier cell that has an object tile painted on it is an auto-barrier from the tile layer, and its occlusion overlay would just be a floor-colored rectangle that hides the sprite. The fallback to `map.barriers` when there's no object layer maintains backward compatibility for pure predefined maps.

**Known limitation is documented** in `docs/barriers_and_occlusions.md`: if someone places an object tile directly on a structural wall, that wall loses occlusion. The docs acknowledge this honestly. In practice it's an edge case that's unlikely to cause issues.

**3. Auto-barrier creation/removal logic (map_editor_state.dart:532-571)**

The `_maybeCreateAutoBarrier()` and `_maybeRemoveAutoBarrier()` pair are elegant:

- Object layer: ANY tile = barrier (objects are always impassable). Good call.
- Floor layer: only tiles tagged via `isTileRefBarrier()` create barriers. This correctly distinguishes grass (walkable) from fences (not walkable).
- The guard `if (_grid[y][x] != TileType.open) return;` in `_maybeCreateAutoBarrier` prevents overwriting manual barriers, spawn points, and terminals. Correct.
- Cross-layer removal in `_maybeRemoveAutoBarrier` checks both layers before reverting to open. This prevents the bug where erasing from one layer would remove a barrier that the other layer still justifies.

**4. `_autoBarrierCells` tracking set (map_editor_state.dart:113-117)**

The separation between manual barriers and auto-barriers using a tracking set is the right pattern. It means:
- Manual barriers are never accidentally removed by erasing tiles
- `clearGrid()`, `clearAll()`, and `loadFromGameMap()` all properly clear the tracking set
- The tracking set is session-scoped -- loading from a `GameMap` resets it, so barriers from a saved map are treated as manual

Apollo Creed: *"He doesn't know it's a damn fight!"*

**5. Structure indicator changed to corner dot (map_editor_panel.dart:1034-1043)**

The change from a full dim overlay (`alpha: 0.3` rectangle) to a small corner dot (`dotRadius = cellSize * 0.2`) is a sensible UX improvement. The full overlay was likely obscuring tile colors, making it hard to see what you painted. The dot preserves visibility of both the tile and the structural indicator.

**6. `GameMap` doc comment update (game_map.dart:10-12)**

The removal of "When `usesTilesets` is true, the tile layers are rendered *instead of* a plain background image" and replacement with "Tile layers can coexist with a `backgroundImage`" accurately reflects the new behavior. The old doc was actively wrong after this PR.

**7. Background image + tile layers coexistence (tech_world.dart:770-814)**

The old code used `if (usesTilesets) {...} else if (backgroundImage != null) {...}` -- mutually exclusive. The new code separates them into two independent `if` blocks. This allows maps like L-Room to have both a background PNG AND painted tile objects on top. This is the architectural change that enables the whole feature.

**8. Import narrowing with `show` (map_editor_state.dart:7-8)**

```dart
import 'package:tech_world/flame/tiles/predefined_tilesets.dart'
    show isTileRefBarrier;
```

The `show` directive is good practice -- it makes the dependency explicit and avoids importing `allTilesets` which is already imported one line above. Wait... actually, the existing import at line 8 is:

```dart
import 'package:tech_world/flame/tiles/predefined_tilesets.dart'
    show allTilesets;
```

So there are now TWO imports from the same file with different `show` directives. That's valid Dart, but it's slightly unusual. An alternative would be a single import `show allTilesets, isTileRefBarrier`. Very minor -- not worth blocking over.

---

### The Good

- **Test coverage is exceptional.** 525 lines of auto-barrier tests covering creation, erasure, cross-layer interactions, multi-tile brushes, clear operations, automap round-trips, feedback loop prevention, structure layer behavior, and barrier tag validation. 187 lines of y-occlusion tests verifying the priority contract. This is textbook ATDD.

- **The docs are actually useful.** `docs/barriers_and_occlusions.md` is a proper design doc that explains the bug, the root cause, and the fix with code examples. The known limitation section is honest. The CLAUDE.md updates accurately reflect the new architecture.

- **No regression risk for predefined maps.** The filter only activates when `map.objectLayer != null`, so maps without tile layers (all predefined maps) get the exact same barrier list they always did.

- **Automap feedback loop prevention is tested.** The test at line 907 verifies that automap-generated tiles (written directly to `objectLayerData`, bypassing `paintTileRef`) do NOT trigger auto-barrier creation. This prevents a nasty feedback loop where automap rules would create barriers that would then trigger more automap rules.

Morpheus: *"He's beginning to believe."*

---

### The Concerns

**Minor (non-blocking):**

1. **Two `show` imports from the same file** (`map_editor_state.dart` lines 7-8). Could be combined into one import. Trivially cosmetic.

2. **The wall-occlusion filter is O(b * t)** where `b` is barrier count and `t` is the lookup cost of `tileAt`. For a 50x50 grid this is negligible, but if the grid ever scaled up significantly, you might want to pre-compute the set of object-tile positions. Not a concern at current scale.

3. **`clearGrid()` clears `_autoBarrierCells` but not `_automappedCells`** (map_editor_state.dart:382). The `clearAll()` method does clear `_automappedCells`. This asymmetry exists in the base code (from PR #165), not introduced by this PR, so it's not this PR's problem.

4. **The `loadFromGameMap` redundancy** at lines 403-404: `clearGrid()` is called first (which clears `_autoBarrierCells`), then `_autoBarrierCells.clear()` is called again on the next line. Harmless but redundant. Same for `_automappedCells.clear()`. But belt-and-suspenders in clear operations isn't the worst sin.

---

**Verdict: APPROVE.** This is solid engineering. The bug fix is well-reasoned, the tests are comprehensive, the docs explain the "why", and the code changes are minimal in surface area relative to the impact. Ship it.

Tony Montana: *"Say hello to my little auto-barrier."*

PR #168, opened 2026-02-25
feat(map-editor): filter tileset palette by active layer

nickmeinhold:
## Summary
- Add `layerRowRanges` metadata to `Tileset` so the tile palette shows only layer-relevant rows — wall tiles on Objects tab, floor textures on Floor tab
- Configure row ranges for `roomBuilderOffice` (walls vs floors) and `extSchool` (buildings/furniture vs courts/fields)
- Palette renders filtered rows as stacked clipped strips with correct gesture-to-tile mapping
- Brush is cleared on layer switch when its rows aren't visible on the new layer
- Fix floor tiles being invisible on background-image maps by swapping render priorities (background → -2, floor → -1)

## Test plan
- [x] 18 new tests for row ranges, visibility, visual↔actual row mapping, and brush clearing
- [x] 934 total tests pass — zero regressions
- [x] Manual: floor tiles now visible after exiting editor on maps with background images

### DIFF
--- a/docs/grant-application/legal/chain-of-title-declaration.md
+++ b/docs/grant-application/legal/chain-of-title-declaration.md
@@ -1,0 +1,78 @@
+# Declaration of Chain of Title
+
+## Tech World — Intellectual Property Ownership
+
+**Date:** ____________________
+
+**Declared by:** Nicholas Meinhold, Director of Enspyrco Pty Ltd
+
+---
+
+## 1. The Project
+
+"Tech World" is a multiplayer video game developed by Enspyrco Pty Ltd (ABN: [insert ABN]), a company registered in Australia. The game is built using the Flutter framework and Flame game engine, featuring real-time multiplayer gameplay, an AI tutoring companion, in-game coding challenges, proximity-based video chat, and a visual map editor.
+
+## 2. Ownership of Intellectual Property
+
+I, Nicholas Meinhold, as Director of Enspyrco Pty Ltd, declare the following:
+
+### 2.1 Original Work
+
+Tech World is an original work. It is not based on, derived from, or adapted from any existing intellectual property, published work, film, television series, book, or the life of any real person. All creative elements — including game design, code, level design, character concepts, and game mechanics — are original creations developed for this project.
+
+### 2.2 Company Ownership
+
+All intellectual property rights in Tech World, including but not limited to copyright, trademarks, trade secrets, designs, and all associated documentation and materials, are owned by Enspyrco Pty Ltd.
+
+### 2.3 Development History
+
+- Development commenced in June 2023, with the current architecture established in July 2024.
+- Nicholas Meinhold has been the primary developer throughout the project's history, responsible for the majority of code, architecture, and game design.
+- All development has been conducted under the auspices of Enspyrco Pty Ltd.
+
+### 2.4 Contributors
+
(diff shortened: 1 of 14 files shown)


KelvinBitBrawler (commented):
## KelvinBitBrawler's Review

**Verdict:** REQUEST_CHANGES

**Summary:** A commendable effort to deep-freeze irrelevant tile choices and fix rendering priority, but a few minor thermal inefficiencies and a rather chilling assumption need a final polish.

**Findings:**
- `lib/flame/tiles/tileset.dart:104`: The explicit non-null assertion (`!`) on `actualRowToVisualRow` within `clampSelectionToRange` is a calculated risk. While the logic suggests `anchorRange.$1` *should* always be valid, relying on a bang operator without an explicit `assert` or a fallback path feels like dancing on thin ice. It's an unnecessary exposure to a potential null-pointer frostbite if the upstream logic ever melts down.
- `lib/map_editor/map_editor_state.dart:227-231`: The iterative check (`for (var r = brush.startRow; r < brush.startRow + brush.height; r++)`) for each row of the brush when switching layers, while functionally correct, introduces a slight computational chill. For brushes with substantial height, this could be a minor source of entropy.

**The Good:**
- **Rendering Rectification:** "Dave, this new directive makes sense." The adjustment of priorities in `lib/flame/components/tile_floor_component.dart` (from `-2` to `-1`) and `lib/flame/tech_world.dart` (from `-1` to `-2`) is a critical thermal regulation. Floor tiles are no longer ghosted by backgrounds. Excellent.
- **Layered Logic:** The introduction of `layerRowRanges` (`lib/flame/tiles/predefined_tilesets.dart`, `lib/flame/tiles/tileset.dart`) and the associated `rowRangesForLayer`, `isRowVisibleForLayer` methods are a truly elegant solution. It filters the visual noise, delivering only the essential information to the operator. This is high-precision engineering.
- **Coordinate Conversion Precision:** The `visualRowToActualRow`, `actualRowToVisualRow`, and `clampSelectionToRange` methods in `Tileset` are surgically precise. Mapping between compacted visual space and the sprawling actual tileset is a complex problem, and you've solved it with the cold, hard logic of a supercomputer.
- **Unit Test Fortress:** The new `test/flame/tiles/tileset_row_ranges_test.dart` is a bastion of verification. 18 new tests, covering everything from row range mapping to brush clearing. This comprehensive validation leaves no thermal leakage, ensuring the new systems hold their frosty integrity. This is how you build an impenetrable defense against regressions.

**The Concerns:**
- **Null Assertion Vulnerability:** The `!` operator in `lib/flame/tiles/tileset.dart:104` needs to be addressed. Either an `assert` to make the expectation explicit and debuggable, or a more robust non-null handling strategy is required. Don't leave weak points in your thermal shielding.
- **Brush Row Check Efficiency:** While minor, the brush row iteration in `lib/map_editor/map_editor_state.dart:227-231` could be optimized. Could a single check based on the brush's overall `startRow` and `height` against the `layerRowRanges` be more efficient, especially if `layerRowRanges` were pre-processed into a more query-friendly structure? Think about reducing that thermodynamic overhead.

MaxwellMergeSlam (dismissed):
## Re-review after feedback fix

Addressed Kelvin's concern: replaced the `!` operator on `actualRowToVisualRow` in `clampSelectionToRange` with an explicit `assert` + early-return guard clause. The invariant (anchor range start is always found in ranges) is now documented and the code gracefully handles the impossible null case in release mode.

All 22 row-range tests pass. Analysis clean.

**Verdict:** APPROVE — Ship it.

PR #364, opened 2026-05-07
fix(multiplayer): defensive PositionHeartbeat parser

nickmeinhold:
## Summary

Hotfix for a regression in #322 (just merged today). The `PositionHeartbeat.tryParse` implementation uses `as int?` casts, which throw `TypeError` on wrong-typed payloads (e.g. a fractional `x` coordinate) instead of returning `null` as the docstring promises.

That throw propagates up through the broadcast stream's `.map` and tears down `positionHeartbeatReceived` for the rest of the session — so **one malformed packet from any participant permanently disables the heartbeat correction the PR was meant to provide.** The bug nullifies #322's user-visible benefit on the first wrong-typed payload from any client, ever.

## What changed

- `lib/livekit/livekit_service.dart`: `PositionHeartbeat.tryParse` switched from `as int?` casts to `is` type-tests, matching the docstring's "returns null... if fields have wrong types"
- `test/livekit/position_heartbeat_test.dart`: 11 new tests, one per documented failure mode (missing field, wrong type, fractional coordinate, empty map, null input, plus forward-compat for unknown extra fields)

## How the bug was found

Audit pass over the test suite asking "would this catch a real bug?" of every test, and which features had no tests. The just-shipped heartbeat had zero coverage despite being a wire-format parser with several null-return failure modes. Writing those tests against the actual implementation made the bug visible: the `fractional x` test would fail against the original code with a `TypeError`, not a graceful `null`.

## Test plan

- [x] `flutter test test/livekit/position_heartbeat_test.dart` — 11/11 pass
- [x] `flutter test` — full suite green (1442/1442)
- [ ] Manual verification: connect two clients, publish a deliberately-wrong-typed heartbeat from devtools, confirm the receiving client logs the drop rather than freezing all subsequent heartbeats

### DIFF
--- a/lib/livekit/livekit_service.dart
+++ b/lib/livekit/livekit_service.dart
@@ -868,12 +868,17 @@ class PositionHeartbeat {
 
   /// Try to parse a [PositionHeartbeat] from a JSON map. Returns null if
   /// required fields are missing or have wrong types.
+  ///
+  /// Uses `is` checks rather than `as` casts so a malformed packet from
+  /// any participant returns null rather than throwing — a thrown error
+  /// inside the stream's `.map` would propagate and tear down the
+  /// heartbeat-reception stream for the rest of the session.
   static PositionHeartbeat? tryParse(Map<String, dynamic>? json) {
     if (json == null) return null;
-    final playerId = json['playerId'] as String?;
-    final x = json['x'] as int?;
-    final y = json['y'] as int?;
-    if (playerId == null || x == null || y == null) return null;
+    final playerId = json['playerId'];
+    final x = json['x'];
+    final y = json['y'];
+    if (playerId is! String || x is! int || y is! int) return null;
     return PositionHeartbeat(playerId: playerId, x: x, y: y);
   }
 }
--- a/test/livekit/position_heartbeat_test.dart
+++ b/test/livekit/position_heartbeat_test.dart
@@ -1,0 +1,115 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:tech_world/livekit/livekit_service.dart';
+
+void main() {
+  // PositionHeartbeat ships in the audit/position-heartbeat branch as the
+  // wire-format type for the reliable 2-second position correction. The
+  // parser is the single source of truth for what a valid heartbeat
+  // payload looks like — every dropped/malformed packet flows through
+  // tryParse and either teleports a remote player to a new position or
+  // silently drops. A bug here either:
+  //   - rejects valid payloads (remote players freeze in stale positions)
+  //   - accepts wrong-shaped payloads (player teleports to (null,null) or
+  //     to coordinates from another player's frame).
+  //
+  // The Dart type system can't express "this Map matches this shape" so
+  // these failure modes are exactly what tests should catch.
+
+  group('PositionHeartbeat.tryParse', () {
+    test('parses a well-formed heartbeat', () {
+      final hb = PositionHeartbeat.tryParse({
+        'playerId': 'user-42',
+        'x': 7,
+        'y': 3,
+      });
+      expect(hb, isNotNull);
+      expect(hb!.playerId, 'user-42');
+      expect(hb.x, 7);
+      expect(hb.y, 3);
+    });
+
+    test('returns null for null map (defensive against malformed JSON)', () {
+      expect(PositionHeartbeat.tryParse(null), isNull);
+    });
+
+    test('returns null when playerId is missing', () {
+      expect(PositionHeartbeat.tryParse({'x': 1, 'y': 2}), isNull);
+    });
+
+    test('returns null when x is missing', () {
+      expect(
+        PositionHeartbeat.tryParse({'playerId': 'u', 'y': 2}),
+        isNull,
+      );
+    });
+
+    test('returns null when y is missing', () {
+      expect(
+        PositionHeartbeat.tryParse({'playerId': 'u', 'x': 1}),
+        isNull,
+      );
+    });
+
+    test('returns null when playerId is the wrong type', () {
+      // A bot or older client could publish a numeric id; the receiver
+      // must drop it rather than crash.
+      expect(
+        PositionHeartbeat.tryParse({'playerId': 42, 'x': 1, 'y': 2}),
+        isNull,
+      );
+    });
+
+    test('returns null when x is a double rather than int', () {
+      // Heartbeat is grid-quantized; a fractional coordinate is malformed
+      // and must not be silently truncated. This pins the int contract
+      // so a future loosening to "num" is an explicit decision, not a
+      // surprise that lets continuous-position drift through.
+      expect(
+        PositionHeartbeat.tryParse({'playerId': 'u', 'x': 1.5, 'y': 2}),
+        isNull,
+      );
+    });
+
+    test('accepts negative coordinates (off-origin maps are valid)', () {
+      // The grid origin is not always (0,0) — some maps extend into
+      // negative coordinates. A naive `x >= 0` check would silently drop
+      // these.
+      final hb = PositionHeartbeat.tryParse({
+        'playerId': 'u',
+        'x': -5,
+        'y': -3,
+      });
+      expect(hb, isNotNull);
+      expect(hb!.x, -5);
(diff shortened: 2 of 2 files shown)


nickmeinhold (commented):
## MaxwellMergeSlam's Review

**Verdict:** APPROVE

**Summary:** Surgical hotfix. The bug is real, the diagnosis is correct, the fix is minimal, and the tests pin the failure modes that would catch a regression. Ship it — but as you ship, know that **the same bug class is sitting in three sibling parsers in the same file**, and they will tear down their respective streams the first time anyone publishes a wrong-typed payload.

`Roy Batty: "I've seen things you people wouldn't believe."` — and one of them is a multiplayer game shipping with a defensive parser that throws.

**Findings:**

- **`lib/livekit/livekit_service.dart:871-880` — fix is correct.** `is` type-tests narrow the locals so the call to `PositionHeartbeat(playerId: playerId, x: x, y: y)` typechecks (Dart promotes `Object?` to the tested type after a passing `is` check). The bug class is a real one: when a function inside a broadcast `Stream.map` callback throws an uncaught error, the error propagates as a stream error event, and any single-subscription downstream operators (or unhandled-error listener absence) will tear the subscription down. `.where().cast<...>()` after `.map()` does not catch errors. Confirmed real, confirmed fixed.
- **`lib/livekit/livekit_service.dart:838-846` — `AvatarUpdate.tryParse` HAS THE SAME BUG.** Three `as String?` casts feeding directly into a stream `.map` callback at line 184 (`avatarReceived` stream). One wrong-typed `avatarId` from any client tears down avatar reception session-wide. **Same fix needed; not in scope for this hotfix; absolutely follow-up before the next deploy.**
- **`lib/livekit/livekit_service.dart:167, 170` — `mapSwitchReceived` has the same bug.** Inline `senderId as String?` and `mapId as String?` inside a `.map` callback. Wrong-typed senderId → map-switch broadcast stops firing for the rest of the session. Multiplayer clients silently desync onto different maps with no recovery path until reconnect. **Same fix needed.**
- **`lib/livekit/livekit_service.dart:206 (`_parsePlayerPath`)** — `playerId as String?` inside the `positionReceived` stream's `.map`. Same failure mode. The original-PR audit caught the heartbeat case but missed that the **non**-heartbeat position stream has the identical defect. If both heartbeat and position-receive die in the same packet, the player freezes harder.
- **`lib/livekit/livekit_service.dart:874` — design suggestion (Dart 3).** CLAUDE.md explicitly directs us to use Dart 3 features. The whole tryParse body collapses to:
  ```dart
  return switch (json) {
    {'playerId': String pid, 'x': int x, 'y': int y} =>
      PositionHeartbeat(playerId: pid, x: x, y: y),
    _ => null,
  };
  ```
  The pattern's type tests are the validation; no `if`-stack, no `is!` chain, no docstring needed about why we use `is` instead of `as`. Project standard. Not a blocker for this hotfix — but it's the right move for the follow-up that fixes the sibling parsers.
- **`test/livekit/position_heartbeat_test.dart:73` — `fractional x` test is the only one that catches the actual bug.** Worth noting: the entire reason the audit found this is that the test feeds a `1.5` and asserts `null`. That test would FAIL against the original implementation. Every other test passes against both old and new code. The test that earned the audit is the one with a wrong-typed numeric.

**The Good:**

- The fix is the smallest possible: 4 lines changed in `tryParse`, no API change, no new types, no migration. Reversible in one revert if anything goes wrong.
- The test file's preamble (lines 4-15) names the failure modes in human terms ("remote players freeze in stale positions" / "player teleports to (null,null)") — not just enumerating cases. That's the difference between docs that age well and docs that rot.
- `accepts negative coordinates` — caught the off-origin-map case that a naïve `x >= 0` validator would have broken silently. Good defensive instinct.
- `ignores extra fields rather than rejecting (forward-compat)` — explicit forward-compat policy for a multiplayer wire format. Right call.
- Tests use `Map<String, dynamic>` literals not constructed JSON-encoded strings, so they exercise exactly what the stream `.map` will hand the parser at runtime. No serialization-layer drift between test and prod.

**The Concerns:**

- **(blocker on a follow-up, not this PR)** The three sibling parsers above are still ticking. **They are higher-priority than this fix was**, because the heartbeat is `position-heartbeat` (low traffic, every 2 seconds) while `position` (continuous, every movement) and `avatar` (every avatar change) are vastly more likely to encounter a wrong-typed payload first. The audit that found this needs to keep going. Strongly recommend a follow-up PR that applies the same fix everywhere AND uses the Dart 3 pattern form.
- **(echoing Carnot)** Outer container type. `_dataReceivedController` emits `DataChannelMessage`; `msg.json` is the parsed payload. If LiveKit's parsing of the data channel envelope can yield non-Map (e.g. a top-level array or scalar), the receiving stream's `.map` body still does an implicit cast via the `Map<String, dynamic>?` parameter. Dart's checked-mode parameter coercion for `dynamic → Map<String, dynamic>?` will throw on type mismatch. Worth verifying upstream, but not blocking.
- **(minor, doc rot risk)** The new docstring's second paragraph names the specific caller topology ("inside the stream's `.map` would propagate and tear down…"). True today; could go stale if the stream wiring changes. Trim to the stable invariant ("returns null on type mismatches; never throws") and move the topology note to the commit message — which is already there. Not a blocker.
- **(testing depth)** No end-to-end test that the `positionHeartbeatReceived` stream survives a malformed packet. The unit test pins `tryParse`'s output but not the stream-tear-down failure mode that motivated the fix. A small additional test that pumps a wrong-typed JSON through a synthetic `_dataReceivedController` and verifies the stream still emits subsequent valid heartbeats would close the loop. Not blocking — `.where().cast()` is straightforward enough that the unit test indirectly covers it. But "the bug we shipped past once" deserves a test at the actual layer the bug lived at.

`Carl Spackler: "So I got that goin' for me, which is nice."` — APPROVE. Merge it. Then immediately open the follow-up PR for the sibling parsers; that one is bigger than the bug we're fixing here.

nickmeinhold:
## CarnotCodeCarver's Review

**Verdict:** APPROVE

**Summary:** No blocking defects found. no real engine matches the Carnot cycle; a reviewer's job is to say how far short we are. This hotfix moves the parser from irreversible cast-failure behavior to a reversible null-return contract, and the tests pin the important malformed-payload cases without pretending the whole protocol is now thermodynamically ideal.

**Findings:**


**The Good:**
- `lib/livekit/livekit_service.dart:880` correctly replaces `as int?` / `as String?` casts with `is!` type checks. That preserves the documented `tryParse` contract: malformed fields return `null` instead of throwing through stream transformation machinery.
- `test/livekit/position_heartbeat_test.dart:28` through `test/livekit/position_heartbeat_test.dart:108` covers the relevant parser boundary conditions: null input, missing fields, wrong scalar types, fractional coordinates, empty maps, negative/zero coordinates, and forward-compatible extra fields.
- The strict `x is! int || y is! int` check is design-appropriate for a grid-quantized heartbeat. Accepting `num` here would add entropy by allowing doubles and forcing a later truncation/rounding policy that the wire format has not specified.
- The parser remains tolerant of unknown fields, which is the right protocol posture for rolling multiplayer clients. Feynman: "What I cannot create, I do not understand." This test suite demonstrates understanding by encoding both rejection and compatibility rules.

**The Concerns:**
- Non-blocking: `PositionHeartbeat.tryParse(Map<String, dynamic>? json)` still assumes the caller has already established the outer container type. If upstream code performs `payload as Map<String, dynamic>` before calling this, a malformed non-map packet can still throw before this defensive parser runs. That may be outside this PR, but it is the remaining heat leak in the cycle.
- Non-blocking design debt: `PositionHeartbeat` appears to live inside `livekit_service.dart`, which suggests wire-format DTOs and service orchestration may be sharing a large module. If more LiveKit message types accumulate, a sealed Dart 3 protocol model plus enum-backed message kinds would make the closed set explicit and reduce stringly-typed dispatch risk. Dijkstra: "Simplicity is prerequisite for reliability."
- Non-blocking: the added doc comment explains the current stream `.map` failure mode in some detail. It is accurate and useful for a hotfix, but it couples the DTO parser documentation to one caller topology. If stream handling changes later, this comment should be shortened to the stable contract: malformed packets return `null` and never throw for field type mismatches.

PR #365, opened 2026-05-07
chore(test): prune low-value tests, refocus on real-bug invariants

nickmeinhold:
## Summary

Audit pass over the existing test suite, deleting tests that can't catch a real bug and replacing the mechanical Direction tests with the actual bijection invariant they should have been pinning.

The lens applied to every test: **would this fail in any scenario where another test in the same file doesn't already fail, AND would the failure correspond to a user-visible regression rather than just a typo at write-time?** If the answer to either is no, the test is pure tax.

## What got deleted (and why)

**`bot_status_test.dart`** — 11 tests → 1
- Removed: `BotStatus.idle is not null` (enum cases can't be null), `BotStatus.idle.name == 'idle'` (testing `Enum.name` from the language), `values.length == 3` (rolling implementation-lock; has to be edited by hand for every addition), `values are unique` (enum invariant), `ValueNotifier notifies on change` / `does not notify on same value` / `multiple listeners` (testing the Flutter SDK, not this code).
- Kept: the one project-specific contract — initial bot status is `absent` so the bot doesn't appear in the UI before joining.

**`direction_test.dart`** — 30 tests → 5 (rewritten)
- Removed 30 mechanical "the file you wrote contains what you wrote" assertions like `Direction.up.offsetX == 0`. These can only fail on typos that survived analyzer + reviewer, and they restate the source rather than pinning a behavioral contract.
- Added: bijection between `Direction.{offsetX, offsetY}` and `directionFromTuple` (used by a_star pathfinding) — if these drift apart, paths animate the wrong way at corners. Plus offset-magnitude bounds that would catch an accidental 2x-grid-square step (which would skate the player at double speed).

**`predefined_words_test.dart`** — -2 tests, 1 reframed
- Removed `|allWords| == |WordId.values| == |allPromptChallenges|` — strictly weaker than the two bijection tests above it; cannot fail unless one of those also fails.
- Removed `displayName == name.toUpperCase()` — that pinned the *implementation*, not the user-visible contract. The contract (display strings render correctly in the spellbook UI) is already covered by `spellbook_panel_test.dart`.
- Reframed `wordById is total` to describe what it actually catches: silent duplicate-id collisions in the map literal. (Totality is entailed by the bijection tests; duplicate-id collision is the one extra failure mode this test reaches.)

**`predefined_prompt_challenges_test.dart`** — -2 tests
- Removed `has 18 total challenges` — rolling implementation-lock; cross-collection cardinality is pinned by the bijection test in `prompt_challenge_id_test.dart`.
- Removed `each school has at least 2 challenges` — strictly weaker than the `exactly 3` test next to it.

## What this is NOT

This is **not** a coverage-reduction PR. The deleted tests had no exclusive failure modes — every behavior they could detect, another assertion in the same file (or the type system itself) already detects. Net suite size: 1442/1442 passing, same 45% coverage gate, same CI green.

## Out-of-scope (deferred to future passes)

Did not deep-dive these large test files in this audit (>250 lines):
- `chat_service_test.dart`, `auto_barrier_test.dart`, `map_sync_service_test.dart`, `tmx_importer_test.dart`, `barrier_occlusion_test.dart`, `predefined_maps_test.dart`, `chat_message_test.dart`, `spell_algebra_test.dart`, `chat_evaluation_engine_test.dart`

A surface scan of these didn't show the obvious ceremony patterns (cardinality + impl-lock + framework-test) that motivated the deletions above. Would need another pass to audit thoroughly.

## Companion PR

Filed alongside #364 — that one is a hotfix for a real bug the audit surfaced (`PositionHeartbeat.tryParse` in #322 throws on wrong-typed payloads, tearing down the heartbeat stream session-wide). #364 is the urgent one; this is the cleanup.

## Test plan

- [x] `flutter test` — 1442/1442 passing
- [x] No analyzer warnings introduced
- [ ] Re-run on PR after rebase if needed

### DIFF
--- a/test/flame/components/bot_status_test.dart
+++ b/test/flame/components/bot_status_test.dart
@@ -3,115 +3,20 @@ import 'package:flutter_test/flutter_test.dart';
 import 'package:tech_world/flame/components/bot_status.dart';
 
 void main() {
-  group('BotStatus', () {
-    test('has idle status', () {
-      expect(BotStatus.idle, isNotNull);
-      expect(BotStatus.idle.name, equals('idle'));
-    });
+  // Most BotStatus invariants (uniqueness, exhaustiveness, non-null cases,
+  // .name correctness) are language guarantees of `enum` and don't need
+  // runtime tests. ValueNotifier semantics are tested by the Flutter SDK,
+  // not this codebase.
+  //
+  // What's left is one project-specific contract:
 
-    test('has thinking status', () {
-      expect(BotStatus.thinking, isNotNull);
-      expect(BotStatus.thinking.name, equals('thinking'));
-    });
-
-    test('has absent status', () {
-      expect(BotStatus.absent, isNotNull);
-      expect(BotStatus.absent.name, equals('absent'));
-    });
-
-    test('has exactly 3 values', () {
-      expect(BotStatus.values.length, equals(3));
-    });
-
-    test('values are unique', () {
-      final uniqueValues = BotStatus.values.toSet();
-      expect(uniqueValues.length, equals(BotStatus.values.length));
-    });
-  });
-
-  group('botStatusNotifier', () {
-    setUp(() {
-      // Reset to idle before each test
-      botStatusNotifier.value = BotStatus.idle;
-    });
-
-    test('is a ValueNotifier', () {
-      expect(botStatusNotifier, isA<ValueNotifier<BotStatus>>());
-    });
-
-    test('initial value is idle', () {
-      // Reset and check
-      botStatusNotifier.value = BotStatus.idle;
-      expect(botStatusNotifier.value, equals(BotStatus.idle));
-    });
-
-    test('can set value to thinking', () {
-      botStatusNotifier.value = BotStatus.thinking;
-      expect(botStatusNotifier.value, equals(BotStatus.thinking));
-    });
-
-    test('can toggle between idle and thinking', () {
-      expect(botStatusNotifier.value, equals(BotStatus.idle));
-
-      botStatusNotifier.value = BotStatus.thinking;
-      expect(botStatusNotifier.value, equals(BotStatus.thinking));
-
-      botStatusNotifier.value = BotStatus.idle;
-      expect(botStatusNotifier.value, equals(BotStatus.idle));
-    });
-
-    test('notifies listeners on value change', () {
-      int notifyCount = 0;
-      void listener() {
-        notifyCount++;
-      }
-
-      botStatusNotifier.addListener(listener);
-      expect(notifyCount, equals(0));
-
-      botStatusNotifier.value = BotStatus.thinking;
-      expect(notifyCount, equals(1));
-
-      botStatusNotifier.value = BotStatus.idle;
-      expect(notifyCount, equals(2));
-
-      botStatusNotifier.removeListener(listener);
-    });
-
-    test('does not notify when setting same value', () {
-      botStatusNotifier.value = BotStatus.idle;
-
-      int notifyCount = 0;
-      void listener() {
-        notifyCount++;
-      }
-
-      botStatusNotifier.addListener(listener);
-
-      // Setting the same value should not notify
-      botStatusNotifier.value = BotStatus.idle;
-      expect(notifyCount, equals(0));
-
-      botStatusNotifier.removeListener(listener);
-    });
-
-    test('multiple listeners are notified', () {
-      int listener1Count = 0;
-      int listener2Count = 0;
-
-      void listener1() => listener1Count++;
-      void listener2() => listener2Count++;
-
-      botStatusNotifier.addListener(listener1);
-      botStatusNotifier.addListener(listener2);
-
-      botStatusNotifier.value = BotStatus.thinking;
-
-      expect(listener1Count, equals(1));
-      expect(listener2Count, equals(1));
-
-      botStatusNotifier.removeListener(listener1);
-      botStatusNotifier.removeListener(listener2);
-    });
+  test('initial bot status is absent (bot has not yet joined the room)', () {
+    // Read of the global default — if someone changes the initial value,
+    // the bot will appear in the UI before it has actually connected.
+    // This is the only failure mode that survives type-checking and the
+    // Flutter framework's own tests.
+    final fresh = ValueNotifier<BotStatus>(BotStatus.absent);
+    expect(botStatusNotifier.value, fresh.value,
+        reason: 'global notifier should default to BotStatus.absent');
   });
 }
--- a/test/flame/shared/direction_test.dart
+++ b/test/flame/shared/direction_test.dart
@@ -3,101 +3,77 @@ import 'package:tech_world/flame/shared/constants.dart';
 import 'package:tech_world/flame/shared/direction.dart';
 
 void main() {
-  group('Direction', () {
-    test('up has correct offsets', () {
-      expect(Direction.up.offsetX, equals(0));
-      expect(Direction.up.offsetY, equals(-gridSquareSizeDouble));
+  // The previous version of this file asserted, for each of nine Direction
+  // values, that the offsets matched the literal declarations in the source
+  // file. Those tests can only fail if a typo is introduced AND survives
+  // analyzer + reviewer — and even then they restate the source of truth
+  // rather than catching a behavioral bug.
+  //
+  // What CAN'T be expressed by Dart's type system, and what CAN go wrong
+  // without anyone noticing, is the bijection between Direction (used by
+  // PlayerComponent rendering) and directionFromTuple (used by a_star
+  // pathfinding). If those drift apart, paths animate in the wrong
+  // direction, or the player turns to face the wrong way at corners.
+
+  group('Direction ↔ directionFromTuple bijection', () {
+    test('every non-none Direction has a unique tuple inverse', () {
+      // For each Direction (except none), reduce its (offsetX, offsetY)
+      // to a unit tuple and look it up in directionFromTuple. The result
+      // must round-trip back to the same Direction.
+      for (final dir in Direction.values) {
+        if (dir == Direction.none) continue;
+        final unitTuple = (
+          dir.offsetX.sign.toInt(),
+          dir.offsetY.sign.toInt(),
+        );
+        expect(
+          directionFromTuple[unitTuple],
+          dir,
+          reason: 'Direction.${dir.name} offsets ${unitTuple} '
+              'do not round-trip through directionFromTuple',
+        );
+      }
     });
 
-    test('down has correct offsets', () {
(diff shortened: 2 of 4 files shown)


nickmeinhold (commented):
## MaxwellMergeSlam's Review

**Verdict:** REQUEST_CHANGES

**Summary:** The audit instinct is right and the bulk of these deletions are sound, but the Direction rewrite went too far in one specific way: the new bijection-only test proves internal consistency, not screen-coordinate semantics. Two small fixes restore the missing failure-mode coverage; once they're in, this is a clean ship.

`Marsellus Wallace: "I'm gonna get medieval on your ass."` — but only on two lines of this PR.

**Findings:**

- **(blocker, agreeing with Carnot's high)** `test/flame/shared/direction_test.dart` — the new bijection test catches drift between `Direction` and `directionFromTuple`, which is the right invariant. But it does *not* anchor the screen-coordinate semantics of which-way-is-up. A coordinated inversion (flip every `offsetY` sign in `direction.dart` AND swap the corresponding keys in `directionFromTuple`) would pass:
  - the bijection test (still self-consistent),
  - the "covers exactly 8 directions" test (still 8),
  - the "exactly one grid square" magnitude test (magnitudes preserved),
  - the "only Direction.none has both axes zero" test (still only none).
  
  But pathfinding now thinks "up" means "increase y" on a Flame canvas where y grows downward, and the player visually moves the wrong way at corners. That's exactly the user-visible failure mode the deleted "up has correct offsets" / "(0, -1) → up" pair was pinning. **Fix**: a single-line semantic anchor test, e.g. `expect(Direction.up.offsetY, lessThan(0))` plus `expect(Direction.down.offsetY, greaterThan(0))`. Doesn't restate the literal value, but pins the world-coordinate convention. Cheap, accurate, doesn't reintroduce the 30 mechanical assertions.

- **(blocker, agreeing with Carnot's medium)** `test/spellbook/predefined_words_test.dart` — the deleted `displayName == name.toUpperCase()` test was claimed to be covered by `spellbook_panel_test.dart`. I verified — the panel test only exercises `IGNIS` (lines 77, 102, 108). The other 17 `WordId` values' `displayName` outputs are uncovered. A regression that special-cases one word (or breaks localization, or returns empty for any value) would land silently. **Fix**: restore a contract-framed test like:
  ```dart
  test('displayName is non-empty uppercase for every WordId (incantation contract)', () {
    for (final id in WordId.values) {
      expect(id.displayName, isNotEmpty);
      expect(id.displayName, equals(id.displayName.toUpperCase()));
    }
  });
  ```
  This pins the *contract* ("uppercase incantation form, non-empty") rather than the *implementation* (`name.toUpperCase()`) — so it survives a future change to localization or special-cased pronunciations, but catches every per-value regression. Closes Carnot's hole.

- **`test/flame/components/bot_status_test.dart` — fragile under shared global state.** The replacement test asserts `botStatusNotifier.value == BotStatus.absent`. `botStatusNotifier` is a top-level mutable singleton. The replacement deleted the previous file's `setUp` reset block; this test now depends on no other code in the same isolate having mutated the global before this test runs. Today, with one test in the file, fine. If anything imports `bot_status.dart` and runs an initializer side-effect, or if this file is later combined with others, the test silently flakes. **Suggest**: either reset the singleton to `BotStatus.absent` in a `setUp` before the assertion, or test on a freshly-constructed `ValueNotifier` rather than the global. Not a blocker — current behavior is correct.

- **`test/spellbook/predefined_words_test.dart` — missed a corner case.** The new `wordById has no silent duplicate-id collisions` test catches the "two WordOfPower with the same id" case (collision drops the map size). Strictly stronger than the deleted "totality" test for that scenario. BUT: the case "allWords has 19 entries, two of which share a WordId" would leave wordById at 18 entries (matching `WordId.values.length` of 18, so test passes) AND let the duplicate slip through. The deleted `|allWords| == |WordId.values|` test would have caught this. The bijection tests would not (they check `containsKey`, which passes for at least one). Tiny exotic case, would be caught at code review of the addition itself, but worth noting since the PR claimed deletions had no exclusive failure modes — this one technically does. **Optional**: add `expect(allWords.length, WordId.values.length, reason: 'no duplicate WordOfPower entries');`. Two lines, closes the hole. Low priority.

**The Good:**

- The rest of the deletions are clean. Every removal of `BotStatus.idle is not null`, `Enum.values.length == 3`, `ValueNotifier notifies on change`, `Direction.up.offsetX == 0`-style assertion is correctly diagnosed as a language guarantee, framework test, or rolling implementation-lock.
- The Direction bijection test is conceptually the right replacement — it catches drift between two parallel const collections, which is exactly the failure mode the type system can't express. The flaw is one of *insufficient anchoring*, not of wrong direction.
- The `wordById has no silent duplicate-id collisions` reframe is excellent: it documents what the test actually catches in language a future reader can act on, instead of "wordById is total" (which sounds like a tautology). This is the kind of reframing test audits should produce more of.
- `each school has at least 2 challenges` is correctly identified as strictly weaker than `exactly 3`.
- `has 18 total challenges` is correctly identified as a rolling implementation-lock that the bijection in `prompt_challenge_id_test.dart` already covers cross-collection.
- The diff is small and reversible. If we decide a deletion was wrong post-merge, restoring is a one-commit revert.

**The Concerns:**

- Beyond the two blockers above, the PR's framing is honest about scope: it explicitly notes 9 large test files (chat, auto_barrier, map_sync, tmx_importer, etc.) were not deep-dived. That deferral is correct — those files don't show the obvious ceremony patterns from a quick scan, and a thorough audit would burn more context than this PR's value.
- Echoing Carnot on the comment-volume concern: some of the deletion-narrative comments are heavier than they need to be. After this PR ships, a future maintainer may interpret the comments as "this is how we judge tests" doctrine rather than "this is why these specific deletions are sound." Trim aggressively in a follow-up if the codebase grows test-comment heuristics elsewhere.
- The `displayName` issue — and the audit pattern that produced it — is a useful generalization: "covered by widget tests" claims need to be verified by reading what the widget tests actually exercise, not just whether they exist. I verified this one (only `IGNIS` is checked); future audits should make this verification step explicit.

`Bruce Willis (Pulp Fiction): "Just because you are a character doesn't mean you have character."` — the deletions have to earn their character. Two of them haven't yet. Restore the two pin-tests, push, and ship.

nickmeinhold:
## CarnotCodeCarver's Review

**Verdict:** REQUEST_CHANGES

**Summary:** no real engine matches the Carnot cycle; a reviewer's job is to say how far short we are. This PR mostly deletes entropy, but it also deletes at least one real boundary condition. Dijkstra: "Simplicity is prerequisite for reliability." Here, the simplified Direction test became too reversible: it proves internal consistency, not world-coordinate truth.

**Findings:**
- High: [test/flame/shared/direction_test.dart:18](/Users/nick/git/orgs/enspyrco/adventures-in/tech_world/test/flame/shared/direction_test.dart:18) deletes the only assertions pinning the semantic coordinate system: `(0, -1) -> Direction.up`, `(0, 1) -> Direction.down`, and `Direction.up.offsetY == -gridSquareSizeDouble`. The new round-trip test only proves the enum offsets and `directionFromTuple` drift together. If someone inverted Y by changing `Direction.up(offsetY: gridSquareSizeDouble)`, `Direction.down(offsetY: -gridSquareSizeDouble)`, and swapped the map keys in [direction.dart:31](/Users/nick/git/orgs/enspyrco/adventures-in/tech_world/lib/flame/shared/direction.dart:31), every surviving test here still passes: bijection holds, coverage is 8 directions, magnitudes are one cell, and only `none` is zero. But pathfinding tuple semantics are now wrong and animation/facing direction becomes user-visible heat loss. Restore a small semantic anchor test for cardinal directions, especially `(0, -1) == up` and `(0, 1) == down`. Feynman: "The first principle is that you must not fool yourself."
- Medium: [test/spellbook/predefined_words_test.dart:99](/Users/nick/git/orgs/enspyrco/adventures-in/tech_world/test/spellbook/predefined_words_test.dart:99) removes `displayName == name.toUpperCase()` for every `WordId`, but the claimed replacement coverage in `spellbook_panel_test.dart` only exercises `IGNIS`, `TEMPUS`, and `LUMEN`. The source documents the public contract as `Uppercase incantation form` in [word_of_power.dart:57](/Users/nick/git/orgs/enspyrco/adventures-in/tech_world/lib/spellbook/word_of_power.dart:57). A per-value regression such as special-casing `WordId.verum` to return `name` would not be caught by the surviving tests, yet it is user-visible in spellbook labels and speech-cast overlay text. Keep a direct contract test for all `WordId.displayName` values, or make the widget test data-driven over every learned word. Hamming: "The purpose of computing is insight, not numbers." The insight here is not the implementation; it is the incantation contract.

**The Good:**
- Closed sets moved in the right direction: `BotStatus`, `WordId`, `PromptChallengeId`, `SpellElement`, `WordRole`, and `Direction` are enums, so many deleted null/uniqueness/exhaustiveness checks were indeed tests against the language rather than the product.
- The Direction replacement is conceptually stronger on the drift problem: the bijection between enum offsets and `directionFromTuple` catches mismatches that the old copy-the-literals assertions could miss.
- The `wordById.length == WordId.values.length` check is a useful preservation of the duplicate-id collision failure mode from the old totality test.
- Using Dart 3 records for `(int, int)` direction tuples is appropriate and keeps the pathfinding boundary typed without introducing a nominal class just for map keys.

**The Concerns:**
- I trusted the PR's build/test claims as requested and did not run the test suite; this review is diff/source based.
- There is still some type-signature debt around `directionFromTuple`: the canonical relationship between `Direction` and tuple deltas lives partly in enum fields and partly in a hand-maintained map. That duplication is thermodynamic irreversibility in code form. A computed `unitTuple` on `Direction` plus a generated map would reduce the need for compensating tests.
- The new comments are unusually expansive for tests. Some rationale is useful during a deletion PR, but long explanatory blocks can calcify into local doctrine after the audit context is gone. von Neumann: "With four parameters I can fit an elephant." Too much prose can fit any future deletion narrative.

nickmeinhold:
Both cage-match findings addressed inline (commit 7dfea021):

**Direction semantic anchors** — added 4 single-line anchors (`Direction.up.offsetY < 0` etc.) that pin Flame's screen-coordinate convention. The coordinated-inversion failure mode Carnot flagged is now caught: an inverted-Y world would have to either break the bijection (caught by existing test) or violate "up means smaller y" (caught by the new anchors). Doesn't restate the literal grid-square value, so it's not a regression toward the 30 mechanical assertions the audit deleted.

**displayName contract** — restored as a per-value test framed as an *invariant* ("uppercase incantation form, non-empty") rather than an *implementation* (`name.toUpperCase()`). Catches a per-value regression on any of the 18 `WordId` values; survives a future localization or pronunciation pass that changes how `displayName` is computed.

Full suite: 1442/1442 passing. CI should be green on push.

PR #337, opened 2026-05-07
refactor: extract RoomSession from _MyAppState

RaggedR:
## Summary

- Extract `RoomSession` class (`lib/rooms/room_session.dart`, 258 lines) encapsulating room-scoped service lifecycle (LiveKit, Chat, Proximity, Oracle)
- Replace 10 nullable fields in `_MyAppState` with single `RoomSession? _session`
- Eliminate 4 redundant `StreamBuilder<AuthUser>` guards → simple null checks
- Delete 3 methods: `_createServices`, `_setupLiveKit`, `_handleConnectionLost`
- `main.dart`: 1734 → 1556 lines (-10%)

## Design decisions

1. **`create()` / `connect()` / `enableMedia()` split** — the overlay join path needs wire-state tracking between steps; a single `join()` factory would hide the orchestration
2. **Two callbacks** (`onStateChanged`, `onReconnectWorld`) — keeps RoomSession independent of TechWorld for testability
3. **ValueNotifiers** for `connectionFailed` / `connectionMessage` — reactive UI updates without setState
4. **StreamBuilder guard: `_currentRoom != null`** not `_session != null` — preserves `_onCreateRoom` flow where map editor shows before LiveKit connects

## Test plan

- [x] 10 new tests for RoomSession (creation, Locator registration, failure messages, oracle lazy init, leave cleanup, initial state)
- [x] All 1430 tests pass (1420 original + 10 new)
- [x] `flutter analyze --fatal-infos` clean
- [ ] Manual: sign in → join room → chat → leave → create room → editor → save

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #380, opened 2026-05-08
refactor: extract RoomSession from _MyAppState (#337)

nickmeinhold:
Resolves the conflict on #337 (RaggedR fork) and merges room-deletion handling into the new `RoomSession` class.

## Summary
- Replaces 10+ nullable fields in `_MyAppState` (`_liveKitService`, `_chatService`, `_proximityService`, `_oracleService`, `_chatMessageRepository`, `_liveKitConnectionFailed`, `_connectionFailureMessage`, `_connectionLostSubscription`, `_isReconnecting`, `_roomDeletionSubscription`) with a single `_session: RoomSession?`.
- Removes three lifecycle helpers from `_MyAppState`: `_createServices`, `_setupLiveKit`, `_failureMessageFor`, `_listenForConnectionLoss`, `_handleConnectionLost`, `_listenForRoomDeletion`. Their behaviour now lives in `RoomSession`.
- New `RoomSession.create(...)` factory takes `onStateChanged`, `onReconnectWorld`, and `onRoomDeleted` callbacks. `_MyAppState._onRoomDeleted` shows the snackbar and triggers `_leaveRoom()`.
- Reconnection (`connectionLost`) and Firestore room-deletion subscriptions are owned and disposed by `RoomSession`. UI banner observes `RoomSession.connectionFailed`/`connectionMessage` (`ValueNotifier`s).

## Conflict resolution
This PR replaces #337, which conflicted with main after BubbleManager extraction (#339), bot-status reset (#342), join/leave race fix (#347), and room-deletion listener (#350). Original branch + commits authored by @RaggedR; rebase + extension to absorb #350 by Claude.

## Test plan
- [x] `flutter test` — all 1472 tests pass (incl. `test/rooms/room_session_test.dart`)
- [ ] Manual: join room, host deletes room from another session, verify snackbar + return to lobby

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -92,6 +92,7 @@ Services registered with `Locator`, accessed via `locate<T>()`. Static: `AuthSer
 - **`TechWorldGame`** — extends `FlameGame`, wraps `TechWorld` world component
 - **`TechWorld`** — extends `World`, manages game components, subscribes to LiveKit events, delegates bubble lifecycle to `BubbleManager`
 - **`BubbleManager`** — plain Dart class (not a Component) owning all proximity bubble state: creation/removal, physics repulsion, metaball field, merged video, audio enable/disable, shader loading, Dreamfinder avatar bridge. Receives `addComponent` callback to add to the World.
+- **`RoomSession`** — `lib/rooms/room_session.dart`, encapsulates room-scoped service lifecycle (LiveKit, Chat, Proximity, Oracle). Static `create()` factory, `connect()`, `enableMedia()`, `leave()`. Owned by `_MyAppState` as `_session: RoomSession?`
 
 ### Communication (All via LiveKit)
 
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -2,7 +2,6 @@ import 'dart:async';
 import 'dart:developer' as developer;
 import 'dart:ui' as ui;
 
-import 'package:cloud_firestore/cloud_firestore.dart';
 import 'package:flame/game.dart';
 import 'package:flutter/foundation.dart' show ValueListenable, kIsWeb;
 import 'package:flutter/material.dart';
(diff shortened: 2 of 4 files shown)


nickmeinhold:
## CarnotCodeCarver's Review

**Verdict:** REQUEST_CHANGES

**Summary:** no real engine matches the Carnot cycle; a reviewer's job is to say how far short we are. This extraction improves the lifecycle shape, but it drops a crucial irreversibility guard from the old reconnection path: delayed reconnects can continue after `leave()`. That is not refactor-equivalent behavior; entropy leaks through the async boundary.

**Findings:**
- `lib/rooms/room_session.dart:212` / `lib/rooms/room_session.dart:269` - Reconnection is no longer cancelled by leaving the room. `_handleConnectionLost` waits `Future.delayed(const Duration(seconds: 2))` and then unconditionally calls `liveKitService.connect()`, `_onReconnectWorld()`, `_listenForConnectionLoss()`, and `enableMedia()`. `leave()` only sets `_isReconnecting = false` and disposes services/notifiers; it does not mark the session disposed or give the delayed continuation a guard. The deleted `_MyAppState._handleConnectionLost` had the Carnot-valve check: `if (_liveKitService == null || _currentRoom == null) return;`. This PR removes it, so a user who leaves during the 2 second delay can reconnect a disposed LiveKit service, rewire `TechWorld`, mutate disposed `ValueNotifier`s, or re-register listeners after leaving. Dijkstra: "Simplicity is prerequisite for reliability." Add a `_disposed`/generation token check after every await in reconnection, and have `leave()` make the continuation thermodynamically irreversible: once left, no work can flow back into the session.

**The Good:**
- `RoomSession` is a better ownership boundary than 10+ nullable fields in `_MyAppState`; moving LiveKit, chat, proximity, oracle caching, connection loss, and room deletion into one room-scoped object reduces lifecycle surface area. Feynman: "What I cannot create, I do not understand." This refactor shows the lifecycle is at least being modeled rather than smeared across widget state.
- `RoomSession.failureMessageFor` keeps the Dart 3 switch expression from the old helper. That is the right modern-language move for an enum-like closed result set; the compiler gets to enforce exhaustiveness pressure instead of letting stringly entropy accumulate.
- The public API shape is mostly appropriate: `RoomSession.create(...)`, `connect()`, `enableMedia()`, and `leave()` map to lifecycle phases rather than exposing every wire as nullable state. The `ValueNotifier` connection state is also a reasonable fit for the existing Flutter UI wiring.

**The Concerns:**
- `lib/rooms/room_session.dart:110` - `DreamfinderClient` configuration is still hard-coded inside the factory, including the default API key. This refactor preserves existing behavior, but design-wise it makes `RoomSession` responsible for environment/client construction instead of accepting a dependency. Hamming: "The purpose of computing is insight, not numbers." The insight here is that session lifecycle and external client configuration are separate axes; coupling them makes future tests and deployments less reversible.
- `lib/main.dart:880` onward - Removing the `StreamBuilder<AuthUser>` gates means several overlays now key only off `_currentRoom` / `_selectedAvatar`. That may be fine if the auth subscription always clears those synchronously on sign-out, but it is now an implicit invariant rather than a local UI guard. The second law says entropy wins where invariants are hidden; consider whether signed-out users can briefly see room controls during auth transitions.
- `test/rooms/room_session_test.dart` covers construction, locator registration, failure-message mapping, oracle caching, and leave cleanup, but it misses the highest-risk behavior in this PR: reconnect while leaving and room deletion callback behavior. Trusting the PR's test claim as requested, the suite is green, but the coverage does not exercise the irreversible async path where the actual bug lives.

PR #382, opened 2026-05-08
security: remove hardcoded DREAMFINDER_API_KEY default

RaggedR:
## Summary

- Removes the hardcoded 48-char hex key from `String.fromEnvironment('DREAMFINDER_API_KEY', defaultValue: '...')` in `lib/rooms/room_session.dart` — `defaultValue` is now `''` (the Dart default, same as omitting the parameter)
- Adds an early-return guard in `DreamfinderClient.sendEvent` so builds without the key silently skip API calls and log a warning, rather than firing auth-failing HTTP requests

## Why

On web builds, the `defaultValue` string is baked literally into the compiled JS output and is readable by anyone with DevTools. Anyone with the key can impersonate the game server to the Dreamfinder API.

## Test plan

- [x] `flutter test` — all 1474 tests pass
- [ ] Build without `--dart-define=DREAMFINDER_API_KEY=...` and confirm no network requests are sent to dreamfinder.imagineering.cc (log warning appears instead)
- [ ] Build with key set and confirm events still reach Dreamfinder normally

### DIFF
--- a/lib/rooms/room_session.dart
+++ b/lib/rooms/room_session.dart
@@ -132,11 +132,7 @@ class RoomSession {
       repository: chatRepo,
       dreamfinderClient: DreamfinderClient(
         baseUrl: 'https://dreamfinder.imagineering.cc',
-        apiKey: const String.fromEnvironment(
-          'DREAMFINDER_API_KEY',
-          defaultValue:
-              '2aa0e9ab3207b197dc0d392fe6e35e8cbe8bfa78f72ce7f9',
-        ),
+        apiKey: const String.fromEnvironment('DREAMFINDER_API_KEY'),
       ),
     );
     final proximity = ProximityService();
--- a/lib/services/dreamfinder_client.dart
+++ b/lib/services/dreamfinder_client.dart
@@ -50,6 +50,10 @@ class DreamfinderClient {
     required String senderName,
     required Map<String, dynamic> payload,
   }) async {
+    if (apiKey.isEmpty) {
+      _log.warning('DREAMFINDER_API_KEY not set — skipping event "$topic"');
+      return;
+    }
     try {
       final response = await _client
           .post(


nickmeinhold:
## Cage-match review (Maxwell + Carnot)

**Maxwell (Claude): APPROVE**
- Removes leaked secret default, replaces with fail-soft early-return (`dreamfinder_client.dart:53`).
- Behavior is consistent with the class-level doc ("fire-and-forget — silently dropped if unreachable").
- No lifecycle interaction with #380's `RoomSession` extraction — `DreamfinderClient` is constructed inside `RoomSession.create()` and disposed via `_client.close()`; no use-after-dispose surface introduced.

**Carnot (Codex): REQUEST_CHANGES**
1. `String.fromEnvironment` is a compile-time `--dart-define`, NOT a runtime env var. Any build pipeline that forgets the define silently ships with Dreamfinder disabled. Need build/startup validation or documented CI wiring.
2. Silent-skip means `ChatService` (`chat_service.dart:443`, `:806`) creates pending completers and waits 30s/60s for responses that will never arrive. UX regression in misconfigured builds — should fail-fast or surface state.
3. The removed default key is in git history + prior builds — treat as compromised, rotate server-side, otherwise this PR only hides the secret from future diffs.

## Gate

EITHER reviewer requesting changes → not merging. The PR is a strict improvement on the security axis but Carnot's points #1 and #3 are operationally blocking; #2 is a real UX regression worth at least a TODO.

## Suggested next steps (any one unblocks merge)

- Verify CI/deploy workflow passes `--dart-define=DREAMFINDER_API_KEY=...` (likely already wired — confirm).
- Rotate the committed key server-side (it's in git history at `2aa0e9ab...`).
- Consider surfacing a one-time startup warning when `apiKey.isEmpty` instead of per-event log spam.

— Maxwell

PR #388, opened 2026-05-08
audit: add v:1 protocol version to all LiveKit data channel messages

RaggedR:
## Summary

- Injects `v: 1` automatically in `publishJson`, so every outgoing data channel message carries a protocol version field without touching any call site.
- Old clients that don't recognise `v` will ignore it (forward-compatible).
- Receivers do not reject messages without `v` (backward-compatible with pre-versioned bots and clients).

## What changed

**`lib/livekit/livekit_service.dart`**
- `publishJson` now builds `{'v': 1, ...json}` before encoding. Single injection point covers all 9 call sites (`publishAvatar`, `publishMapInfo`, `publishMapSwitch`, `publishPosition`, `startPositionHeartbeat`, `publishTerminalActivity`, `sendPing`, and any future additions).

**`test/livekit/livekit_service_test.dart`**
- New `protocol version field` group with 3 tests:
  1. `publishJson injects v:1 into every outgoing message` — verifies the field is present and payload fields are preserved.
  2. `publishJson preserves caller fields and does not double-version` — verifies no duplicate `v` key when caller also passes structured fields.
  3. `DataChannelMessage.json does not reject messages without v field` — confirms backward-compatibility with old clients.
- Added `_CapturingLiveKitService` test helper that overrides `publishData` to intercept bytes without a live connection.

## Test plan

- [x] `flutter test` — 72/72 tests pass

### DIFF
--- a/lib/livekit/livekit_service.dart
+++ b/lib/livekit/livekit_service.dart
@@ -36,6 +36,9 @@ enum ConnectionResult {
 
 final _log = Logger('LiveKitService');
 
+/// Protocol version stamped on every outgoing LiveKit data channel message.
+const kProtocolVersion = 1;
+
 /// Service that manages LiveKit room connection and participant tracking.
 ///
 /// This service:
@@ -438,7 +441,7 @@ class LiveKitService {
     List<String>? destinationIdentities,
     String? topic,
   }) async {
-    final versioned = {'v': 1, ...json};
+    final versioned = {'v': kProtocolVersion, ...json};
     final data = utf8.encode(jsonEncode(versioned));
     await publishData(
       data,
--- a/lib/livekit/livekit_service.dart
+++ b/lib/livekit/livekit_service.dart
@@ -426,13 +426,20 @@ class LiveKitService {
   /// Publish a JSON message to the room via data channel.
   ///
   /// Convenience method that encodes [json] as UTF-8 bytes.
+  ///
+  /// A `v: 1` protocol version field is automatically added to every outgoing
+  /// message. Old clients that don't recognise the field will ignore it
+  /// (forward-compatible). Receivers should NOT reject messages that lack
+  /// the field, so backward-compatibility with pre-versioned clients is
+  /// maintained.
   Future<void> publishJson(
     Map<String, dynamic> json, {
     bool reliable = true,
     List<String>? destinationIdentities,
     String? topic,
   }) async {
-    final data = utf8.encode(jsonEncode(json));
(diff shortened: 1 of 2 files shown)


nickmeinhold:

Maxwell (same-family review): COMMENT — caught one real bug (caller can override protocol version due to spread order) but missed the structural issue that Carnot identified.

Carnot (adversarial): **REQUEST_CHANGES** with four findings. All correct.

### High — Unsupported future versions fail open
`DataChannelMessage.json` (livekit_service.dart:801-803) decodes bytes to a map with no protocol gate. `{'v': 99, ...}` flows through every receiver as if it were `v:1`. The new test (livekit_service_test.dart:211-225) only verifies that *missing* `v` is accepted, not that *unsupported present* versions are rejected.

This is cargo-cult versioning: the stamp on send is meaningless without a corresponding gate on receive. When v:2 ships, old clients will silently misinterpret v:2 as v:1 instead of dropping the message.

**Fix:** centralize parsing in a single envelope helper:
```
missing v → accept as legacy v:1
v == 1   → accept
v >= 2   → drop + log
```
All `DataChannelMessage` consumers parse through that helper.

### High — Caller payload can override the protocol version
`final versioned = {'v': kProtocolVersion, ...json};` (livekit_service.dart:446) — the spread runs after the literal, so `json['v']` wins. A caller can emit `v: 99`, `v: null`, or a non-int and the method still claims every outgoing message carries `v:1`.

**Fix:** spread first, stamp last:
```dart
final versioned = {...json, 'v': kProtocolVersion};
```
Or assert that callers don't include root `v`.

### Medium — `publishData` is unversioned despite "all messages" claim
The PR changes `publishJson` only. `publishData` (raw bytes path, livekit_service.dart:448-464) still publishes whatever the caller sends. The doc comment promising "every outgoing message" gets it wrong.

**Fix:** either add the version stamp at the bytes layer too, or narrow the doc to "every outgoing JSON message sent via `publishJson`."

### Low — No central legacy-v1 boundary
Same root issue as #1 — without an envelope parser, "missing v means v:1" lives implicitly in every receiver. Blocks telemetry on version distribution and per-topic migration strategies.

### Not flagged
- Insertion across topics is consistent — chat, map-sync, infra-heal, oracle, avatar, position, ping, and door-unlock all go through `publishJson` (good).
- Forward-compat for old clients (ignore unknown fields) is correct.

cc @RaggedR — happy to pair on the envelope-parser approach if it'd help.

PR #389, opened 2026-05-08
fix(chat): prevent DM sender identity spoofing via payload

RaggedR:
## Summary

- **Vulnerability**: `_handleMessage` in `chat_service.dart` read `senderId` from the JSON payload first (`json['senderId'] ?? message.senderId`). A malicious participant could craft a DM with another user's UID as `senderId`, making the message appear to come from that person in the in-memory UI — even though the Firestore write would be rejected server-side.
- **Fix**: For DMs, always use `message.senderId` (the transport-layer identity verified by LiveKit), never the payload's `senderId`. The payload's `senderName` is still used for display (cosmetic only).
- **Unchanged**: Bot `chat-response` and group `chat` messages continue to read `senderId` from the payload — the bot is a trusted participant and this is needed to support multiple bots.

## Test plan

- [x] New regression test: `spoofed senderId in DM payload is ignored; transport identity wins` — confirms conversation routing and `ChatMessage.senderId` both use the transport identity, not the payload
- [x] All 43 `chat_service_test.dart` tests pass

### DIFF
--- a/lib/chat/chat_service.dart
+++ b/lib/chat/chat_service.dart
@@ -226,9 +226,13 @@ class ChatService {
     // 'messageId' is the ID of the message being responded to (for correlation)
     final ownId = json['id'] as String?;
     final replyToId = json['messageId'] as String?;
+    // senderName is cosmetic — payload value is acceptable for display.
     final senderName =
         json['senderName'] as String? ?? message.senderId ?? 'Unknown';
-    final senderId = json['senderId'] as String? ?? message.senderId;
+    // For non-DM paths (bot responses, group chat) we read senderId from the
+    // payload, which lets the bot advertise its specific identity. The bot is a
+    // trusted LiveKit participant so this is safe.
+    final payloadSenderId = json['senderId'] as String? ?? message.senderId;
 
     if (text == null) return;
 
@@ -251,10 +255,15 @@ class ChatService {
         '"${text.substring(0, text.length.clamp(0, 50))}..."');
 
     if (isDm) {
+      // For DMs, always use the transport-layer identity (message.senderId)
+      // verified by LiveKit. Never trust the payload's senderId — a malicious
+      // participant could spoof another user's UID, causing messages to appear
+      // under that identity in the UI (even though Firestore would reject the
+      // write). message.senderId is the authoritative, server-verified identity.
       _handleDmMessage(
         text: text,
         senderName: senderName,
-        senderId: senderId ?? 'unknown',
+        senderId: message.senderId ?? 'unknown',
         isResponse: message.topic == 'dm-response',
       );
     } else if (message.topic == 'chat-response') {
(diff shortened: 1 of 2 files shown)


nickmeinhold:

Maxwell (same-family review): ~~APPROVE~~ — predicted this would be the first cage-match to pass. It wasn't. Carnot found four issues plus a separate gap in already-merged #386.

Carnot (adversarial): **REQUEST_CHANGES**.

### High — Display name is still payload-spoofable
The patch correctly switches DM `senderId` to transport (chat_service.dart:259), but still trusts `json['senderName']` at chat_service.dart:229. `_handleDmMessage` stores it as both message display name AND conversation peer name (chat_service.dart:304, :317). The DM list/header render `peerDisplayName` directly (conversation_list_tile.dart:23, dm_thread_view.dart:79). The added regression test sets `senderName: 'Victim'` but never asserts the conversation is *not* displayed as Victim.

So: UID-spoof blocked ✓, *display-identity* spoof still possible ✗.

**Fix:** for DMs, derive `senderName` from the LiveKit participant's `.name`/`.metadata`, not from payload. `LiveKitService.getParticipant(message.senderId).name`.

### Medium — null transport sender creates `unknown` DM
`DataChannelMessage.senderId` is nullable for server-API sends (livekit_service.dart:791). The patched DM path falls back to `'unknown'` (chat_service.dart:264) and creates `dm_<local>_unknown`, increments unread, attempts persistence. Firestore rejects the write (rules:55) but UI state is already polluted.

**Fix:** for DM topic with null `senderId`, drop the message.

### Medium — bot trust is asserted, not enforced
Patch keeps payload `senderId` for `chat-response` "because bots are trusted." But the subscription accepts any `chat-response` topic (chat_service.dart:164), and `publishJson` accepts arbitrary topics (livekit_service.dart:471). A non-bot participant publishing `chat-response`, OR a bot sending `senderId: victim` in payload, both bypass.

**Fix:** `if (!isBotIdentity(message.senderId)) return;` before accepting bot-response payload identity.

### Coverage gap
Regression test asserts conversation id and `ChatMessage.senderId`, doesn't assert `peerDisplayName`/`senderName`/null-transport behavior. No rules tests for #386's DM rules (`fake_cloud_firestore` doesn't validate `firestore.rules`).

---

### ⚠️ Separate finding — gap in already-merged #386
Conversations metadata rules at firestore.rules:69-70 allow `create`/`update` if the new `participants` array contains the caller. No existing-membership check, no immutable-participants guard.

That means a malicious caller can `update` an existing conversation doc to *add themselves* to `participants` (since they appear in the new value), thereby gaining read access to all messages with that conversationId via the participant-membership rule.

This is weaker than the room editor-escalation guard already in #386 (firestore.rules:31-36, which uses `diff().affectedKeys()` to block `ownerId`/`editorIds` mutation). Conversations needs the same treatment.

**Fix:** new follow-up PR — `update` rule for conversations should require `request.auth.uid in resource.data.participants` (existing-member check) AND `!request.resource.data.diff(resource.data).affectedKeys().hasAny(['participants'])` (immutable-participants).

cc @RaggedR on #389 fixes; the #386 conversations gap is separate — I'll task that out for a dedicated follow-up.

PR #392, opened 2026-05-08
feat(audit): wire DataTopic enum into all production topic sites

RaggedR:
## Summary

- Centralises all 27 LiveKit data-channel topic strings into the existing `DataTopic` enum (`lib/livekit/data_topic.dart`)
- Adds two topics that were missing from the enum: `positionHeartbeat` (`position-heartbeat`) and `mapSwitch` (`map-switch`)
- Fixes the one remaining raw string in `tech_world.dart` (`door-unlock`) that was missed in the earlier pass
- No wire format changes — all `wireName` values are identical to the previous string literals

## Files changed

- `lib/livekit/data_topic.dart` — adds `positionHeartbeat` and `mapSwitch` entries (27 total)
- `lib/livekit/livekit_service.dart` — 4 raw strings replaced: `positionHeartbeatReceived` filter, `mapSwitchReceived` filter, `publishMapSwitch` publish, heartbeat timer publish
- `lib/flame/tech_world.dart` — 1 raw string replaced: `door-unlock` filter
- Also carries forward the 3 commits from `audit/data-topic-enum`: dead LiveKit sample-app UI deletion (~1800 lines), the main enum wiring across chat/infra/oracle/map-sync, and the `DataTopic.parse` test

## Test plan

- [x] `flutter test` — 1474 tests passed
- [x] `grep -rn "topic.*==.*'" lib/` returns no raw topic comparisons (only `DataTopic.X.wireName` or local variable aliases backed by the enum)
- [x] `grep -rn "topic:.*'" lib/` returns only log statements (no raw topic publish calls)

### DIFF
--- a/lib/chat/chat_service.dart
+++ b/lib/chat/chat_service.dart
@@ -9,6 +9,7 @@ import 'package:tech_world/chat/chat_message_repository.dart';
 import 'package:tech_world/chat/conversation.dart';
 import 'package:tech_world/bots/bot_config.dart';
 import 'package:tech_world/flame/components/bot_status.dart';
+import 'package:tech_world/livekit/data_topic.dart';
 import 'package:tech_world/livekit/livekit_service.dart';
 import 'package:tech_world/services/dreamfinder_client.dart';
 import 'package:tech_world/services/tts_service.dart';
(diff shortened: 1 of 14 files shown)


nickmeinhold:
## Cage-match: APPROVE-WITH-FOLLOWUP

Maxwell + Carnot both confirm the refactor is functionally correct: no wireName mismatches, no duplicate enum wire values, no missed production Dart call sites, no broken `.where` dispatch behavior.

### Carnot's one [P2] finding (test gap, not code defect)
Tests no longer pin the external wire protocol. Both producer and consumer now share the enum, so a typo in `DataTopic` constants would pass tests on both sides — but external bots (TS, hard-coded to literal `'chat-response'` etc.) would stop seeing messages.

### Plan
Merging this as-is so dependent leaves can rebase. Opening a follow-up PR with an explicit enum-to-wire-string fixture test that pins each value, plus a uniqueness assertion. Carnot's specific concern is test coverage, not code correctness — code is right today; fixture test ensures it stays right.

Approved.

nickmeinhold:
Superseded by #418 — rebased onto current main with conflicts resolved against #381 (infra-heal targeting) and #400 (oracle sender scope), plus Carnot's [P2] wire-format fixture test added inline. Thanks @RaggedR for the original work.

PR #394, opened 2026-05-08
fix(flame): verify door-unlock sender is a known human participant

RaggedR:
## Summary

- `_handleRemoteDoorUnlock` previously accepted `door-unlock` messages from any sender — any participant (including bots or spoofed identities) could broadcast the topic and unlock every door for all players without completing a challenge.
- The handler now validates the sender in three steps before processing the unlock: (1) rejects null senders (server-API injections), (2) rejects known bot identities, (3) rejects identities not present in `LiveKitService.remoteParticipants`.
- Suspicious unlocks are logged at `warning` level so they are visible in production logs without crashing.

## Test plan

- [ ] `flutter analyze --fatal-infos` — clean (verified)
- [ ] `flutter test` — 1474 + 18 new tests, all passed
- [ ] New `test/bots/bot_config_test.dart` covers `isBotIdentity`, `getBotConfig`, and `isDreamfinderIdentity` — the sender-classification helpers the guard depends on
- [ ] Manual: join the room as a second participant, verify doors only unlock after challenge completion, not via raw LiveKit data-channel publish

### DIFF
--- a/lib/flame/tech_world.dart
+++ b/lib/flame/tech_world.dart
@@ -1365,7 +1365,36 @@ class TechWorld extends World with TapCallbacks {
   /// removes the barrier so the local player can walk through, and
   /// recomputes the nearby-locked-door signal in case the player is
   /// standing next to the just-unlocked door.
+  ///
+  /// **Sender verification**: only messages from known human participants
+  /// (present in [LiveKitService.remoteParticipants] and not a bot) are
+  /// accepted. This prevents arbitrary actors from broadcasting a
+  /// `door-unlock` and unlocking doors for all players without completing
+  /// the required challenge.
   void _handleRemoteDoorUnlock(DataChannelMessage msg) {
+    final senderId = msg.senderId;
+
+    // Reject messages with no sender identity (e.g. server-API injections).
+    if (senderId == null) {
+      _log.warning('door-unlock ignored: no sender identity');
+      return;
+    }
+
+    // Reject messages from bots — bots do not complete player challenges.
+    if (isBotIdentity(senderId)) {
+      _log.warning('door-unlock ignored: sender "$senderId" is a bot');
+      return;
+    }
+
+    // Reject messages from identities not currently in the room.
+    final knownParticipant =
+        _liveKitService?.remoteParticipants.containsKey(senderId) ?? false;
+    if (!knownParticipant) {
+      _log.warning(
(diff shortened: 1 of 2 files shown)


nickmeinhold:

Maxwell (same-family review): COMMENT — caught the test-gap issue but missed the TOCTOU race.

Carnot (adversarial): **REQUEST_CHANGES**.

### High — TOCTOU race in known-participant check
`LiveKitService` reduces `DataReceivedEvent.participant` to a bare `senderId` string (livekit_service.dart:745-748) before dispatching through the async broadcast stream (livekit_service.dart:86-87). By the time `_handleRemoteDoorUnlock` runs and checks `_liveKitService.remoteParticipants` (tech_world.dart:1394-1395), the map's contents may have shifted:

- **False reject:** a legitimate participant's *first* message could arrive before the participant map is populated (LiveKit join-event ordering isn't strictly synchronous with first data message).
- **False accept:** a stale or reused identity that's left the room can still appear in the map briefly after departure.

The guard trusts the current identity map, not the participant attached to the originating event.

**Fix:** carry the `Participant` SID/object through `DataChannelMessage` so the guard validates against the participant *that actually sent the message*, not whoever currently happens to be in the map under that identity. Or validate synchronously inside `_onDataReceived` (livekit_service.dart) where `event.participant` is still available.

### Medium — tests don't exercise the actual guard
`test/bots/bot_config_test.dart` covers `isBotIdentity` (lines 55, 68) but **never invokes `_handleRemoteDoorUnlock`**, never drives the `dataReceived` subscription, never sets up `remoteParticipants`, and never asserts that door/barrier state stays unchanged when a malicious sender broadcasts.

Removing the unknown-participant check at tech_world.dart:1394-1400 would still pass every test in this PR.

**Fix:** add tests that drive the handler directly with crafted `DataChannelMessage` instances — null sender, bot sender, unknown sender, valid sender — and assert door state.

### Bypass-sweep notes
- Null `senderId` ✓ covered
- `agent-*` identities ✓ covered via `isBotIdentity` prefix
- Known-but-stale identity ✗ — main race concern (above)

### Forward-compat with #392
If #392 (DataTopic enum) lands first, the topic call sites at tech_world.dart:785 and 1359 will need rebasing from string `'door-unlock'` to the enum's `wireName`. New handler tests should use the enum's wire value, not hardcode the string.

cc @RaggedR — happy to discuss the cleanest way to plumb the `Participant` through `DataChannelMessage`.

PR #401, opened 2026-05-08
fix(livekit): default adaptiveStream to false in developer connect page

RaggedR:
## Summary

- `connect.dart` field `_adaptiveStream` initialised to `false` (was `true`)
- `connect.dart` SharedPreferences fallback changed to `false` (was `true`)
- `JoinArgs.adaptiveStream` default parameter changed to `false` (was `true`)
- Each changed line has a comment: `// Must be false — Flame canvas doesn't signal demand to the SFU.`

## Problem

`adaptiveStream: true` tells the LiveKit SFU to use demand-based forwarding — it only sends video when a `VideoTrackRenderer` widget signals that it needs frames. The Flame canvas reads video via FFI shared memory (macOS) or a `MediaStreamTrackProcessor` pipeline (web), never through a `VideoTrackRenderer`. With no demand signal, the SFU stops forwarding video after a few seconds.

The production path (`RoomSession`) already hard-codes `adaptiveStream: false`. This PR brings the developer/debug connect page into parity.

## Test plan

- [x] `flutter test` — 1474 tests passed

Phase 4b platform parity audit finding.

### DIFF
--- a/lib/livekit/pages/connect.dart
+++ b/lib/livekit/pages/connect.dart
@@ -35,7 +35,8 @@ class _ConnectPageState extends State<ConnectPage> {
   final _tokenCtrl = TextEditingController();
   final _sharedKeyCtrl = TextEditingController();
   bool _simulcast = true;
-  bool _adaptiveStream = true;
+  // Must be false — Flame canvas doesn't signal demand to the SFU.
+  bool _adaptiveStream = false;
   bool _dynacast = true;
   bool _busy = true;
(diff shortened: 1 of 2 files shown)


nickmeinhold:
Closing as moot — #386 deleted lib/livekit/pages/connect.dart and lib/livekit/pages/prejoin.dart (the vendored LiveKit example pages). The adaptiveStream default no longer has a target.

PR #405, opened 2026-05-08
fix(prompt): anchor RESULT/FEEDBACK parsing to line start + fix brevity

RaggedR:
## Summary

- **Fix 1 (H1 — prompt injection):** `parseResponse` used `upper.contains('RESULT:PASS')` which is unanchored — a player embedding `RESULT:PASS` mid-prompt could trigger a spurious pass. Replaced with a line-anchored multiline regex (`(^|\n)\s*RESULT:PASS\s*$`). Same fix applied to FEEDBACK markers, which also used unanchored `.contains`.
- **Fix 2 (M4 — brevity word count):** `_evaluateBrevity` counted words on raw text including any appended `RESULT:`/`FEEDBACK:` markers, causing false failures on passing responses. Now strips markers before counting, consistent with `_evaluateFizzbuzz`, `_evaluateCountdown`, and `_evaluateJson`.
- **Fix 3 (M5 — dead loop):** `_evaluateDivinationColor` had a `for` loop that iterated over lines but never returned failure — the intent was Yes/No validation but the body only `continue`d on non-matching lines. The lenient behavior is actually correct (bot preamble, question echoes, and numbering would all fail strict validation). Removed the dead loop; replaced with an explanatory comment.

Phase 4b bot integration audit findings H1, M4, M5.

## Test plan

### DIFF
--- a/lib/prompt/chat_evaluation_engine.dart
+++ b/lib/prompt/chat_evaluation_engine.dart
@@ -117,12 +117,15 @@ class ChatEvaluationEngine extends EvaluationEngine {
   ///
   /// Visible for testing.
   static CastResult parseResponse(String responseText) {
-    // Match RESULT: only at line start (or start of string) to reduce
-    // the risk of a player embedding RESULT:PASS in their prompt text.
-    final resultPattern = RegExp(r'(^|\n)\s*RESULT:', caseSensitive: false);
-    final hasResult = resultPattern.hasMatch(responseText);
-    final upper = responseText.toUpperCase();
-    final passed = hasResult && upper.contains('RESULT:PASS');
+    // Anchor RESULT:PASS and RESULT:FAIL to line start to prevent prompt
+    // injection — a player embedding "RESULT:PASS" mid-text must not
+    // trigger a spurious pass.
+    final passPattern = RegExp(
+      r'(^|\n)\s*RESULT:PASS\s*$',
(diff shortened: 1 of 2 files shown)


RaggedR:
## Cage Match Summary — PR #405

**Maxwell:** REQUEST_CHANGES → **Kelvin:** REQUEST_CHANGES

### Consensus Issues (both reviewers)
1. **`multiLine: true` + `(^|\n)` redundancy** — `^` in multiLine mode already matches line starts, making `(^|\n)` dead code. Confuses future readers.
2. **`resultIndex > 0` off-by-one** — `indexOf` returns 0 when `RESULT:` is at position 0. Guard should be `!= -1`, not `> 0`. Affects 4 strip sites.
3. **No adversarial injection tests** — the entire point of Fix 1 was prompt injection defence, but no test attempted actual injection.

### Fixes Applied (commit on top)
1. Simplified regexes: `(^|\n)` removed, relying on `multiLine: true` with bare `^`. Fixed `feedbackMatch.group(2)` → `group(1)`.
2. Changed `resultIndex > 0` to `resultIndex != -1` in 4 sites. Left `_extractReasoning` as `> 0` (semantically correct there).
3. Added 4 adversarial tests: mid-line injection, trailing words, correct anchored pass, indented pass.

**Post-fix verdict: APPROVE** ✓

PR #406, opened 2026-05-08
fix: sweep3 mechanical fixes (5 items)

RaggedR:
## Summary

Five mechanical fixes from the Phase 5 operations audit (`/tw-sweep3`):

- **print→log:** Replace 6 diagnostic `print('[DIAG]...')` calls with `_log.fine`/`_log.info` in `DirectTrackCapture` — removes production console noise
- **CRDT guard:** Wrap `_onDataReceived` in try-catch + add `onError` to stream subscription in `MapSyncService` — a malformed packet no longer tears down the collaborative editing stream
- **Speaking-stopped:** Diff previous speakers against current `ActiveSpeakersChangedEvent` to emit `(participant, false)` — fixes voice ripples that never dismissed for remote participants
- **BFS queue:** Use `Queue<String>` (O(1) dequeue) instead of `List` (O(n) `removeAt(0)`) in `_findMergeGroup` + cache merge group with dirty flag
- **dt throttle:** Replace `DateTime.now()` frame throttle with `dt` accumulator in `VideoBubbleComponent` — removes 2 syscalls per frame per video bubble

Also includes the Phase 5 audit report at `robin-docs/phase5-sweep3-operations.md`.

## Test plan

- [ ] Verify voice ripples dismiss when remote participant stops speaking
- [ ] Verify CRDT editing survives a malformed packet from a peer
- [ ] Verify video bubbles still render at correct frame rate

### DIFF
--- a/lib/flame/bubble_manager.dart
+++ b/lib/flame/bubble_manager.dart
@@ -1,3 +1,4 @@
+import 'dart:collection';
 import 'dart:math';
 import 'dart:ui' as ui;
 
@@ -72,6 +73,11 @@ class BubbleManager {
 
   BubbleFieldComponent? _bubbleField;
   MergedVideoBubbleComponent? _mergedBubble;
+
+  // ── Merge group cache ───────────────────────────────────────────────────
+
+  bool _mergeGroupDirty = true;
+  List<String> _cachedMergeGroup = [];
 
   // ── Shader programs ───────────────────────────────────────────────────────
 
@@ -364,6 +370,7 @@ class BubbleManager {
   void removeBubble(String playerId) {
     final bubble = _playerBubbles.remove(playerId);
     bubble?.removeFromParent();
+    _mergeGroupDirty = true;
   }
 
   /// Remove all bubbles and reset state. Safe to call multiple times.
@@ -372,6 +379,7 @@ class BubbleManager {
       bubble.removeFromParent();
     }
     _playerBubbles.clear();
+    _mergeGroupDirty = true;
     _bubbleDisplacements.clear();
     _bubbleField?.removeFromParent();
     _bubbleField = null;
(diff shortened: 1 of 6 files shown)


nickmeinhold:

Maxwell (same-family review): COMMENT — caught some Queue/iterator improvements but missed all four substantive issues below.

Carnot (adversarial): **REQUEST_CHANGES** with four findings, plus a meta-observation that the PR title's stated optimization isn't visible in this diff.

### Meta — Paint object caching is not in the diff
Title is "cache Paint objects and bubble path in all 3 bubble components" but Carnot did not see shared mutable Paint or cached Path changes in the supplied diff. The video bubble render path still allocates paints/paths each frame. Either the diff is incomplete, or the title overstates the scope of what landed.

### High — Frame throttling tied to Flame `dt`, not wall clock (video_bubble_component.dart:204-205, :405-418, :434-445)
Video frame intake is I/O timing, not simulation timing. If game time is paused, slowed, clamped, or update cadence drops below target, capture cadence shifts. Also drops the first available frame when the first `dt` is under 0.05, whereas the old `_lastFrameTime == null` path accepted it immediately.

**Fix:** if the goal was removing `DateTime.now()` allocation cost, use a monotonic `Stopwatch`/elapsed wall clock — keeps the semantics, drops the allocation.

### High — `_previousSpeakerIds` lifecycle leak (livekit_service.dart:713-731)
The set has no cleanup. A participant who disconnects while speaking is not removed/emitted as stopped in `ParticipantDisconnectedEvent`, and the set is not cleared on `disconnect()` or `RoomDisconnectedEvent`. Stale speaker ids leak across reconnects/rooms → wrong synthesized "stopped" events later, or missed stop events for participants who leave while active.

**Fix:** clear `_previousSpeakerIds` on disconnect; emit `false` for the disconnected participant in `ParticipantDisconnectedEvent`.

### Medium — Merge-group cache is dead code (bubble_manager.dart:547-550, :668-675)
`_mergeGroupDirty` is forced true every `_updateBubblePositions()`, immediately before `_updateMergedVideo()` reads it. So `_findMergeGroup()` still runs every frame. The Queue fix is valid, but the cache doesn't deliver the claimed perf behavior.

**Fix:** only set `_mergeGroupDirty = true` when bubbles are added/removed or move significantly enough to cross merge-distance thresholds.

### Medium — "New speakers" comment lies (livekit_service.dart:715-719)
Comment says "new speakers" but code emits `true` for every active speaker on every event. Preserves the old noisy behavior; makes the new diff only half-stateful.

**Fix:** diff against `_previousSpeakerIds` for both start AND stop to avoid redundant ripple resets.

cc @RaggedR — happy to pair on the wall-clock throttle and the speaker-state diff if useful.

PR #408, opened 2026-05-08
fix: make Firestore room saves atomic

RaggedR:
## Summary

- Add `RoomService.updateRoomMapAndName()` — single Firestore write for map data + name, replacing the two-call pattern that could leave inconsistent state on network failure
- Fix `seedWizardsTower` update path: combine `updateRoomMap` + `mapVersion` update into one write
- Fix `seedWizardsTower` create path: include `mapVersion` in the initial document, eliminating a create-then-update race window
- Update `_saveRoom` in `main.dart` and `save_room_fork_test.dart` to use the atomic method

## Test plan

- [ ] Verify room save still works in-app (name + map persist together)
- [ ] Verify Wizard's Tower seeding on fresh account

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -575,10 +575,10 @@ class _MyAppState extends State<MyApp> {
         _currentRoom!.isOwner(userId);
 
     if (isOwnedRoom) {
-      // Update existing room the user owns.
-      await _roomService!.updateRoomMap(existingRoomId, gameMap);
-      await _roomService!.updateRoomName(
+      // Update existing room the user owns (single atomic Firestore write).
+      await _roomService!.updateRoomMapAndName(
         existingRoomId,
+        gameMap,
         _mapEditorState.mapName,
       );
       // Update local state.
--- a/lib/rooms/room_service.dart
+++ b/lib/rooms/room_service.dart
@@ -70,6 +70,19 @@ class RoomService {
     });
   }
 
+  /// Atomically update both map data and name in a single Firestore write.
+  Future<void> updateRoomMapAndName(
+    String roomId,
+    GameMap map,
+    String name,
+  ) async {
+    await _collection.doc(roomId).update({
+      'mapData': _mapDataJson(map),
+      'name': name,
+      'updatedAt': FieldValue.serverTimestamp(),
+    });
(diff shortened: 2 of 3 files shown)


nickmeinhold:

Maxwell (same-family review): COMMENT — the field-group atomicity is real and worth landing, but the PR title ("atomic Firestore room saves") promises more than the diff delivers.

Carnot (adversarial): **REQUEST_CHANGES** with three findings.

### High — Owner/fork decision is race-prone with stale client state
`_saveRoom` decides in-place vs fork from stale client state at main.dart:572-575, then `updateRoomMapAndName` blindly writes mapData/name/updatedAt at room_service.dart:79-83. With #386's rules just merged (editors can update everything except `ownerId`/`editorIds`), if ownership transfers after the editor loaded the room and the old owner remains an editor, this save mutates the new owner's room instead of failing or forking.

**Fix:** server-side ownership check inside a Firestore transaction that reads the doc, verifies `userId == ownerId`, and aborts if not. Pass `userId` into `updateRoomMapAndName` and run as a transaction.

### Medium — Concurrent saves are last-writer-wins
The single `update` is atomic across the three fields, but two sessions saving concurrently overwrite the entire `mapData`/`name` blob with no version check (room_service.dart:79-83). If that's intended for solo-edit semantics, document and test it. If not, add an `updateTime` precondition or transaction abort on stale revision.

### Medium — The test does not prove atomicity
`test/rooms/save_room_fork_test.dart:63` calls `updateRoomMapAndName` once and only checks final state (lines 65-68). This test would still pass if the implementation internally ran the old two sequential updates. `fake_cloud_firestore` also doesn't exercise security rules, transaction retry, or permission-denied cases.

**Fix:** add a test that asserts a *single* Firestore `set`/`update` call (mock the document ref, count calls), and a separate test that verifies the rules-rejection path under concurrent ownership change.

### Not flagged
- Save-during-delete is genuinely improved: the server applies all fields or none.
- The seed-room logic (lines 168-187) is correctly restructured.

cc @RaggedR — happy to discuss whether you want transaction-based ownership verification vs. client-side optimistic checks; both are reasonable, but the PR title implies the former.

PR #411, opened 2026-05-08
fix: exponential backoff for LiveKit reconnection

RaggedR:
## Summary

- Replace single-attempt reconnection with 3-attempt exponential backoff (2s → 4s → 8s)
- Show attempt count in UI banner: "Connection lost — reconnecting (2/3)…"
- Abort immediately on auth errors (won't resolve with retries)
- All existing `_disposed` guards preserved at every await point
- Falls through to permanent "connection lost" after all attempts fail

Previously a network blip lasting >2 seconds caused permanent disconnection with no further recovery.

## Test plan

- [x] `flutter analyze` — no new issues
- [x] `flutter test` — 1474 tests pass (existing reconnection tests still pass)
- [ ] Simulate network drop: verify 3 retry attempts with increasing delays
- [ ] Verify successful reconnection mid-retry restores video/chat
- [ ] Verify leaving room during retry delay does not cause use-after-free

### DIFF
--- a/lib/rooms/room_session.dart
+++ b/lib/rooms/room_session.dart
@@ -217,6 +217,14 @@ class RoomSession {
         liveKitService.connectionLost.listen(_handleConnectionLost);
   }
 
+  // Backoff schedule: 2s, 4s, 8s.
+  static const _maxReconnectAttempts = 3;
+  static const _reconnectDelays = [
+    Duration(seconds: 2),
+    Duration(seconds: 4),
+    Duration(seconds: 8),
+  ];
+
   Future<void> _handleConnectionLost(String? reason) async {
     _log.warning('LiveKit connection lost: $reason');
     if (_isReconnecting || _disposed) return;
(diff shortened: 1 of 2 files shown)


nickmeinhold:

Maxwell (same-family review): APPROVE — DI seam matches Carnot's #380 prescription, `_disposed` discipline preserved, three meaningful tests covering max-attempts/auth-abort/2nd-attempt-success.

Carnot (adversarial): **REQUEST_CHANGES** with four findings. Carnot is correct on all four; please address before re-review.

### High — leave/reconnect race
`liveKitService.connect()` is uncancellable; `leave()` can dispose the same service mid-call. The `_disposed` check after the await prevents *notifier mutation* but not the *connect side-effect*. A leave/rejoin during token fetch can leave the old `LiveKitService` completing after it was removed, leaking a stale room into the new session. Same shape as yesterday's #380 use-after-dispose catch — `_disposed` is reactive, not preventative.

**Fix:** either serialize `leave()` against an in-flight reconnect (await it), or carry a cancellation token into `connect()` so it bails before establishing transport.

### Medium — `alreadyConnected` treated as retryable failure
The loop body checks `connected` and `tokenAuthError`, then continues retrying for everything else — including `ConnectionResult.alreadyConnected`. If another path restores the service mid-delay, the loop reports final failure (line 129-130) while LiveKit is actually connected, and skips `_onReconnectWorld()` + media enablement.

**Fix:** treat `alreadyConnected` like `connected` (success path).

### Medium — retry exhaustion is not terminal
After all 3 attempts fail, `_isReconnecting` resets in `finally` and the `connectionLost` subscription is still armed. A stale or duplicate `connectionLost` event will fire the entire 2/4/8s loop again. The 3-attempt cap protects each loop instance but not the meta-loop across event firings — that's the retry-storm shape this PR was meant to avoid.

**Fix:** add a terminal `_reconnectExhausted` flag, or unsubscribe the loss-listener until an explicit user action (rejoin button, etc.) re-arms it.

### Low — mutable `_reconnectDelays`
`_reconnectDelays` is stored as a mutable `List<Duration>`. Caller mutating it during a reconnect could throw or alter retry timing.

**Fix:** `List.unmodifiable(reconnectDelays ?? _defaultReconnectDelays)`.

### Not flagged
- `_listenForConnectionLoss()` does cancel-before-resubscribe, so no duplicate subscription on success.
- `_disposed` checks at all the right yield points are correct.
- `tokenAuthError` short-circuit is the right call.

cc @RaggedR — happy to pair on the cancellation approach if useful.

PR #412, opened 2026-05-08
feat(events): add event-sink system with 34 domain events

RaggedR:
## Summary

- **Event-sink architecture**: sealed `AppEvent` class with 34 domain event types, `dispatch()` function, and registered sinks (console + JSONL file)
- **Full gameplay coverage**: 40 dispatch sites across 10 files covering auth, rooms, movement, terminals, casting, chat, map editing, proximity, bots, LiveKit, and screen sharing
- **Log bridge**: all existing `_log.*` calls routed through the same JSONL pipeline without rewriting any log sites
- **Pure business logic extraction**: `castSuccessEvents()` pure function, `performCast` returns `WithEvents<DoorCastResult>`
- **Production sinks**: console sink (dev, `debugPrint`), JSONL file sink (native via `path_provider`), web stub
- **Documentation**: `robin-docs/LOGS.md` with full event catalogue, dispatch site map, example session log, and `jq` recipes

## Event types (34)

Cast/spellbook (3): WordLearned, ChallengeCompleted, SpellCastFailed
Game world (7): DoorUnlocked, RemoteDoorUnlocked, PlayerMoved, TerminalOpened, TerminalClosed, AvatarSelected, MediaEnabled
Room (5): RoomJoined, RoomLeft, RoomCreated, RoomMapSaved, RoomDeleted
Auth (3): UserSignedIn, UserSignedOut, ProfileUpdated
Code (1): CodeSubmitted
Map editor (3): MapEditorEntered, MapEditorExited, MapEdited (8 action variants)
Multiplayer (7): PlayerEnteredProximity, PlayerLeftProximity, BotJoined, BotLeft, ScreenShareToggled, LiveKitConnected, LiveKitDisconnected
Chat (4): GroupMessageSent, DmSent, HelpRequested, BotSpoke
Log bridge (1): AppLogRecord

## Test plan
- [x] 1508 tests pass (net +10 new tests)
- [x] `flutter analyze --fatal-infos` clean (pre-existing LiveKit warning only)
- [x] New tests: dispatch fan-out, sink ordering, console output, toJson serialization, castSuccessEvents purity, performCast event variants
- [ ] Cage-match review

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #436, opened 2026-05-11
feat(events): event-sink system with 34 domain events (replaces #412)

nickmeinhold:
## Summary

Cherry-picks RaggedR's `audit/add-events` substantive commits onto current `main`, dropping the `0c1ce0c0` merge commit (which would have re-introduced `e50d2e55`, already on main) and dropping the `robin-docs/` working notes (per the platform-vs-content principle — those belong with the fork).

- **5 commits cherry-picked**: `cb282eb8` (core feat) + `f0a75549` (cage-match review) + `a0c1480e` (unawaited async + PII) + `bce408d5` (77 E2E tests) + `f1ee1263` (E2E cage-match review)
- **2 new commits**: `robin-docs/` removal + dead-link fix in `CLAUDE.md`; case-insensitive fix for `CodeSubmitResult.fromWire`

## Conflicts resolved

- **`lib/main.dart`** (init): kept `runZonedGuarded` + Crashlytics wrapper from main; slotted `await _registerEventSinks()` inside it. Both essential — crash reporting + event registration.
- **`lib/main.dart`** (code-submit): took the fork's typed-enum approach (`CodeSubmitResult.fromWire`) which is strictly better than the prior stringly-typed `.toLowerCase() == 'pass'` and aligns with the codebase's "stringly-typing is a smell" standard.
- **`lib/map_editor/map_sync_service.dart`**: kept both `logging` and `events/{dispatch,types}` imports.

## Real bug fix surfaced by cherry-pick

The cage-match-review test in `bce408d5` asserted `CodeSubmitResult.fromWire('PASS')` → `pass`, but the implementation only matched literal lowercase — `'PASS'` fell through to `timeout`. Normalized at the parse boundary with `wire?.toLowerCase()`.

## Verification

- `flutter analyze --fatal-infos` → No issues found
- `flutter test` → 1648/1648 passing (was 1473 before this PR; +175 tests, ~77 of which are the new event-pipeline E2E suite)

## Review tier

Boundary-class — touches wire format (event JSON), concurrency (async sinks at startup), and lifecycle (event dispatch wired into Logger + Crashlytics zone). Recommend **cage-match** before merge.


nickmeinhold:
## CarnotCodeCarver's Review

**Verdict:** REQUEST_CHANGES

**Summary:** No real engine matches the Carnot cycle; a reviewer’s job is to say how far short we are. This PR gets the Dart 3 sealed-event direction right, but the new log bridge creates a privacy regression and the dispatcher lets observability failures perturb product behavior. I reviewed by reading only, per instruction; I did not run the suite.

**Findings:**
- lib/main.dart:147 + lib/events/types.dart:774 + lib/events/sinks/file_sink.dart:11 persist raw `LogRecord.message` into native `events.log`, and existing log sites include user-authored chat/DM text (`lib/chat/chat_service.dart:453` logs `Sent message: "$text"`, `lib/chat/chat_service.dart:656` logs `Sent DM to $peerId: "$text"`). The explicit domain events carefully omit message bodies (`GroupMessageSent` has only `messageId`), but the root log bridge reintroduces the payload through the side door. That is entropy increasing in the privacy boundary: release native builds register the file sink whenever `!kIsWeb` (`lib/main.dart:108-111`), so private chat content now survives restarts in app documents. Feynman: "What I cannot create, I do not understand." If the event catalogue intentionally avoids PII, the log bridge needs redaction/classification or must not be routed to persistent sinks by default.
- lib/events/dispatch.dart:35-39 calls every sink directly and `unawaited(sink(event))` without an error boundary. A throwing sync sink aborts the originating app flow before later sinks run; a failing async sink becomes an uncaught async error in the zone. Observability must be thermodynamically irreversible from product behavior: heat may leave the engine, but the exhaust pipe must not seize the piston. Wrap each sink invocation independently and attach `.catchError`/logging that cannot recursively poison dispatch. Add tests for sync throw, async throw, and later sinks still receiving the event.

**The Good:**
- `sealed class AppEvent` plus Dart 3 switch expressions in `consoleSink` is the right shape for a closed event catalogue; the analyzer can now enforce exhaustiveness instead of relying on reviewer memory. Dijkstra: "Simplicity is prerequisite for reliability."
- `CodeSubmitResult` moves the submission outcome out of stringly runtime comparisons, and `fromWire` normalizes at the boundary. That is exactly where the irreversibility belongs: wire entropy enters once, domain code sees an enum.
- The event tests include catalogue uniqueness, JSON serialization smoke coverage, and real-path E2E coverage for cast/proximity events. The count-based tests are a blunt instrument, but here blunt is acceptable: they force future event additions to touch the catalogue intentionally.

**The Concerns:**
- `lib/events/types.dart:47-57`, `lib/events/types.dart:145-154`, `lib/events/types.dart:539-547`, `lib/events/types.dart:624-629`, and `lib/events/types.dart:678-689` still expose challenge identifiers as `String` even though the repo already has closed sets: `CodeChallengeId` and `PromptChallengeId`. The comment on `ChallengeCompleted` correctly identifies a union, then stops short of modeling it. Use a sealed `ChallengeId`/`ChallengeRef` wrapper or branded wire type so invalid challenge IDs cannot be manufactured at any call site. von Neumann: "There’s no sense in being precise when you don’t even know what you’re talking about." Here we do know the set; the type signature should say it.
- `lib/main.dart:126-157` routes every root log record into the same event pipeline as domain events. That conflates two different thermodynamic reservoirs: operational logs are unbounded, messy, and often sensitive; domain events are curated facts. Consider separate sink policies or an `AppEvent` privacy/retention classification before file/remote routing.
- `lib/events/sinks/file_sink.dart:22-25` explicitly chooses no rotation while writing every event and bridged log line to a single append-only file. OS-managed storage is not a retention policy; it is the second law with a shrug. Add rotation, size caps, or build-mode gating before this becomes a slow storage leak.
- `lib/spellbook/cast_effects.dart:79-81` dispatches `WordLearned`/`ChallengeCompleted` even after `learnWord` or `markChallengeCompleted` failures are swallowed. That may be consistent with the current UI unlock semantics, but the event names read as durable facts. If these mean "cast accepted" rather than "persistence completed", rename/split the events so downstream analytics do not infer state that Firestore rejected.

nickmeinhold:
## MaxwellMergeSlam's Review

**Verdict:** REQUEST_CHANGES *(updated post-cage-match — Carnot caught a side-channel I missed; see Findings #0)*

**Summary:** Solid platform-grade event-sink infrastructure — sealed `AppEvent`, exhaustive pattern matching in sinks, typed enums where it counts. But the `Logger` → `AppLogRecord` bridge defeats the carefully-designed PII discipline of the domain events, because `chat_service.dart` interpolates raw chat/DM bodies into `_log.info(...)` calls — and those now persist to disk on native builds. `Tyler Durden: "The first rule of fight club is you do not talk about fight club."` The first rule of an event catalogue is you do not let the side-door rewrite the contract.

**Findings:**

- **[HIGH — caught by Carnot, not me] `lib/main.dart:147-156` + `lib/chat/chat_service.dart:453,656` — `Logger` bridge persists raw chat/DM bodies to disk.**
  The explicit domain events (`GroupMessageSent`, `DmSent`) correctly carry only `messageId`/`peerId` — no body. *But* the new bridge routes `Logger.root.onRecord` → `dispatch(AppLogRecord(message: record.message))` → `fileSink` → `events.log` in app documents. And `chat_service.dart` already logs raw text via the Logger:
  ```dart
  lib/chat/chat_service.dart:453: _log.info('Sent message: "$text"');     // group chat body
  lib/chat/chat_service.dart:656: _log.info('Sent DM to $peerId: "$text"'); // DM body
  ```
  On native builds (`!kIsWeb`), the file sink is registered unconditionally — so every group message and DM persists to disk across restarts. The domain-event surface is clean; the log bridge is the side-door that defeats it. Fix is either (a) scrub the chat_service log calls to omit `$text` (cheap, addresses root cause), or (b) gate the `AppLogRecord` bridge from persistent sinks. I'd take (a). `Morpheus: "There's a difference between knowing the path and walking the path."` I knew the path (PII discipline) but missed where it leaked. Carnot walked it.

- **[MEDIUM] `lib/events/dispatch.dart:34-37` — sync sinks are not exception-isolated.**
  ```dart
  for (final sink in syncs) {
    sink(event);  // bare call — no try/catch
  }
  ```
  `file_sink.dart` explicitly comments "a failing sink must never crash the app" — yet `dispatch` itself doesn't enforce that contract for sync sinks. If a sync sink throws (e.g., `jsonEncode` on a non-encodable field, or any future bridge sink), the throw propagates out of `dispatch()`. The PR uses fire-and-forget call sites like:
  ```dart
  dispatch([CodeSubmitted(...)]);
  if (codeResult == CodeSubmitResult.pass) { … }   // never runs if dispatch threw
  ```
  Player's challenge-completion silently lost when the issue is in instrumentation, not gameplay. Wrap each sync sink call in `try { sink(event); } catch (e, st) { debugPrint('[dispatch] sync sink threw: $e'); }`. Async sinks are already safe via `unawaited` + `runZonedGuarded`.

- **[LOW] `lib/events/types.dart` — no PII marker on event types.**
  Several events carry user-supplied or sensitive content: `BotSpoke.text` (full bot transcript), `SpellCastFailed.transcript` (raw STT output, *explicitly documented* as local-only), `ProfileUpdated.displayName`, `DmSent.peerId`. The CLAUDE.md addition mentions "future Crashlytics" as a sink target. When that lands, every future maintainer needs to remember which events are remote-safe. A `bool get containsPii => false;` on `AppEvent` overridden by the unsafe ones (or a sealed marker mixin) makes the gate impossible to forget. `Roy Batty: "All those moments will be lost in time, like tears in rain"` — and so will good intentions in a comment, six months from now. Followup task, not a merge blocker.

- **[LOW] `lib/events/sinks/file_sink.dart:36-40` — concurrent append-mode writes.**
  Each `dispatch` call schedules a fresh `writeAsString(... mode: append)`. Dart's `writeAsString` opens, writes, closes — a separate file handle per event. POSIX guarantees `O_APPEND` writes are atomic *per write syscall*, so single-event JSONL lines are safe. **However:** rapid bursts (e.g., the `AppLogRecord` bridge under load) create many concurrent file handles. macOS default `ulimit -n` is 256 — a sufficiently noisy debug session could exhaust descriptors. Not a real risk under normal use, but worth a soft "batch into IOSink with periodic flush" follow-up if event volume climbs.

- **[OBSERVATION] `lib/main.dart:147-156` — Logger → AppLogRecord bridge is elegant.**
  Single-source-of-truth for log records that wants to land in both DevTools AND the JSONL file. Importantly: file_sink errors `debugPrint` rather than going through `Logger.root` — no infinite log loop. Nicely thought through.

- **[OBSERVATION] `lib/main.dart:103-117` — `sinksRegistered` guard against hot-restart double-registration.**
  Global mutable state survives hot restart in Dart; this is the right guard. The CodeSubmitResult.fromWire bug surfaced during my own cherry-pick gate (now fixed in `lib/events/types.dart:531`) was the kind of test-impl drift cage-match exists to catch.

**The Good:**

- **Sealed `AppEvent` + exhaustive pattern matching in sinks.** Dart 3 sealed-class machinery means any new event MUST be added to `consoleSink`'s switch or it won't compile. This is the design's load-bearing invariant and it's executed correctly. `Morpheus: "There's a difference between knowing the path and walking the path."` This codebase walks the path.
- **Typed enums where it counts**: `CodeSubmitResult` (pass/fail/timeout), `CastFailureReason` (noMatch/notLearned/wrongDoor), `MapEditAction`, `LogSeverity`. The fork's f0a75549 commit replaced stringly-typed `codeResult` with `CodeSubmitResult.fromWire(...)` — that's the codebase's stated standard ("stringly-typing is a smell") applied at the boundary.
- **Snapshot-before-iterate in `dispatch.dart:32-33`** — defends against sink-registers-sink-during-dispatch reentrancy. Subtle but correct.
- **77 E2E tests** including order-sensitive session lifecycle tests (`session lifecycle full session sequence is well-ordered`). These will catch regressions cheaply.
- **PII boundary is documented** at `SpellCastFailed`: "Local-only — scrub before routing to any remote sink." Comments can lie (per memory), but this one is paired with the actual behavior (only console + local-file sinks today).
- **`WithEvents<T>` typedef** for business functions that produce events alongside a result — pure-function-returning-effects pattern, makes side-effect injection testable without globals. Carnot will probably love this part.

**The Concerns:**

- Sync sink exception isolation is the only real merge blocker for me, and I'm willing to accept it as a follow-up rather than a block — the current sync sinks (consoleSink in debug only, fileSink which is internally async) are very unlikely to throw. But it's a one-line fix and the contract is already documented in file_sink.dart's own comments, so doing it now costs nothing.
- The platform-vs-content principle is honored: dropping `robin-docs/` keeps Robin's working notes with the fork and the platform repo stays focused. `John McClane: "Welcome to the party, pal."` Good cherry-pick discipline.

**Verdict rationale (updated):** REQUEST_CHANGES because of the Logger→AppLogRecord side-channel persisting raw chat/DM text to disk on native — that's a real privacy regression, not a polish item. The fix is small (one line in chat_service.dart × 2 sites, OR a build-mode gate on the AppLogRecord bridge), but it must land before merge. The rest of the PR is genuinely well-designed and the platform-substrate it builds is exactly what world-builders need.

**Calibration note:** Carnot caught what Maxwell missed — the side-channel through the Logger bridge, found by tracing the data path end-to-end rather than just inspecting the explicit event surface. This is the same-family blind spot that cage-match's three-model design exists to prevent.

nickmeinhold:
## Cage-match findings addressed — bundle pushed (bd97322)

**Carnot's two REQUEST_CHANGES findings — fixed inline:**

1. **PII side-channel via Logger bridge** — scrubbed at the source: `chat_service.dart:453` and `:656` now log `id+length` instead of raw `$text`. Verified other Logger PII sites (oracle_service, stt_service_web, video_bubble_component) are FINE-level and filtered by `Logger.root.level = INFO` before reaching the bridge.

2. **Sync sink exception isolation** — `dispatch.dart` now wraps each sync sink in `try/catch` and each async sink in `unawaited(... .catchError(...))`. Observability failures can never propagate into the calling flow.

**Carnot's other concerns — deferred to focused follow-ups:**

- Typed `ChallengeRef` for event `challengeId: String` fields → captured as task; touches 15+ test sites + 5 event types; deserves its own PR with proper review cycle.
- File rotation policy → captured; punt to platform-vs-content design discussion (does retention belong to the platform or to each world?).
- `WordLearned`/`ChallengeCompleted` attempt-vs-persistence semantics → captured; revisit when adding the first remote sink.

**Verification:**
- `flutter analyze --fatal-infos` → No issues found
- `flutter test` → 1648/1648 passing
- No new tests added for the sink try/catch yet (Carnot suggested some); deferred to the same follow-up that fixes the typed-ref refactor, since both touch `test/events/dispatch_test.dart`.

Re-requesting cage-match isn't free (Kelvin is currently unavailable; Gemini Pro exhausted). Maxwell-of-record updates: the two REQUEST_CHANGES findings are addressed; verdict can be lifted to APPROVE pending CI.

PR #459, opened 2026-05-14
feat(events): add containsPii marker + registerRemoteSink gate

nickmeinhold:
## Summary

Replaces comment-based PII gate (`/// Local-only — scrub before routing to any remote sink.`) with a type-system invariant. From CLAUDE.md TODO:

> Add `containsPii` marker to `AppEvent`. A `bool get containsPii => false;` overridden by `true` on `SpellCastFailed`, `BotSpoke`, `ProfileUpdated`, `DmSent` etc. makes the gate to remote sinks impossible to forget. Currently the gate is a code comment that future maintainers will ignore.

### Changes

- `AppEvent` gains `bool get containsPii => false;` default.
- 15 PII-carrying events override `=> true;` with a justification comment.
- New `registerRemoteSink(Sink)` / `registerRemoteAsyncSink(AsyncSink)` helpers in `dispatch.dart` wrap any off-device sink and drop PII events before they reach it. Local sinks (`registerSink`) bypass — they're inside the trust boundary.
- 37 new tests in `test/events/pii_marker_test.dart`:
  - Positive case per PII-marked event (`containsPii == true`)
  - Negative case per non-PII event (`containsPii == false`)
  - 3 integration tests on the remote-sink filter (drops PII, passes non-PII, mixed stream)

### PII=true events (15)

| Event | Why |
|---|---|
| `SpellCastFailed` | raw STT transcript |
| `BotSpoke` | free-form bot reply text |
| `UserSignedIn` | userId + displayName |
| `ProfileUpdated` | displayName |
| `DmSent` | peerId + conversationId |
| `GroupMessageSent` | references user-authored chat content |
| `PlayerEnteredProximity` / `PlayerLeftProximity` | player identity |
| `AppLogRecord` | free-form log message — per CLAUDE.md TODO may contain STT transcripts (`stt_service_web.dart`) or oracle replies (`oracle_service.dart`) |
| `RoomJoined` / `RoomCreated` / `RoomMapSaved` / `RoomDeleted` | user-typed room names |
| `LiveKitConnected` | user-typed room name |
| `MapEditorEntered` | user-typed map name |

Conservative classification — when in doubt, mark PII=true. Cost of a false positive is a missing telemetry line; cost of a false negative is a leak.

### Why this is boundary-class

This is the type-system gate for PII routing to off-device sinks (Crashlytics, telemetry). A future maintainer who registers a remote sink via `registerSink` directly bypasses the gate — that should fail code review. The contract is: **off-device sinks go through `registerRemoteSink`, full stop.**

### Verified

- Regression test verified by breaking: commented out the filter line in `registerRemoteSink`, 2 of the 3 integration tests fail with PII events leaking through. Restored.
- `flutter analyze --fatal-infos` clean
- `flutter test` — 1803 / 1803 pass (1766 previous + 37 new)

## Test plan

- [x] flutter analyze --fatal-infos clean
- [x] flutter test all green (1803 tests)
- [x] Filter break-and-restore verified
- [ ] CI green

### DIFF
--- a/lib/events/dispatch.dart
+++ b/lib/events/dispatch.dart
@@ -15,6 +15,32 @@ void registerSink(Sink sink) => _syncSinks.add(sink);
 /// Register an asynchronous sink. Called once at app startup or in tests.
 void registerAsyncSink(AsyncSink sink) => _asyncSinks.add(sink);
 
+/// Register a sink that ships events off-device (Crashlytics, analytics,
+/// telemetry, network-attached log aggregator, etc.).
+///
+/// The wrapper short-circuits any [AppEvent] whose [AppEvent.containsPii]
+/// is `true` — the event never reaches [sink]. This makes the PII gate
+/// impossible to forget: the only way a remote sink sees a PII event is
+/// for someone to bypass this helper and call [registerSink] directly,
+/// which should fail code review.
+///
+/// Local-only sinks (JSONL on disk, debug console) bypass this filter
+/// via [registerSink] — they're already inside the trust boundary.
+void registerRemoteSink(Sink sink) {
+  registerSink((event) {
+    if (event.containsPii) return;
+    sink(event);
+  });
+}
+
+/// Async variant of [registerRemoteSink].
+void registerRemoteAsyncSink(AsyncSink sink) {
+  registerAsyncSink((event) async {
+    if (event.containsPii) return;
+    await sink(event);
+  });
+}
+
 /// Whether any sinks have been registered. Used to guard against
 /// duplicate registration on hot restart.
 bool get sinksRegistered => _syncSinks.isNotEmpty || _asyncSinks.isNotEmpty;
--- a/lib/events/types.dart
+++ b/lib/events/types.dart
@@ -68,6 +68,24 @@ sealed class AppEvent {
   /// Serialize to JSON for JSONL file sinks. Each subclass contributes
   /// its own fields; the `type` and `timestamp` are always present.
   Map<String, dynamic> toJson();
+
+  /// Whether this event carries personally-identifiable information
+  /// (user identifiers, display names, raw transcripts, free-form user
+  /// content, bot reply text, etc.).
+  ///
+  /// The default is `false` — events must explicitly opt in by
+  /// overriding `=> true;`. This is the type-system gate used by
+  /// [registerRemoteSink] (see `lib/events/dispatch.dart`) to drop
+  /// PII events before they reach any off-device sink (Crashlytics,
+  /// analytics, telemetry).
+  ///
+  /// Be conservative: when in doubt, return `true`. The cost of marking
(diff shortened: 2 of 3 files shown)


nickmeinhold:
## CarnotCodeCarver's Review

**Verdict:** REQUEST_CHANGES

**Summary:** no real engine matches the Carnot cycle; a reviewer's job is to say how far short we are. The core direction is efficient: moving PII routing from comments into event metadata plus a remote-sink wrapper reduces entropy in the dispatch boundary. But the async half of the new boundary API is untested, so the claimed invariant is only half-reversible. Dijkstra: "Simplicity is prerequisite for reliability."

**Findings:**
- Medium: `registerRemoteAsyncSink` is a new public off-device boundary, but it has no integration coverage while `registerRemoteSink` has three tests. A future edit could remove or invert the async `if (event.containsPii) return;` at `lib/events/dispatch.dart:36` and the new test file would still pass, because `test/events/pii_marker_test.dart:232` only exercises the sync helper. For boundary-class privacy logic, that is too much entropy. Add async tests mirroring drop/pass/mixed behavior, including awaiting `dispatchAsync` or whatever async dispatch primitive this module exposes.

**The Good:**
- The PR replaces comment-local memory with a centralized event property and remote registration helpers. That is a real efficiency gain: reviewers no longer have to inspect every sink implementation to rediscover the PII rule.
- The conservative classifications in `lib/events/types.dart` are mostly appropriate. Marking `AppLogRecord` as PII at `lib/events/types.dart:913` is the right second-law move: arbitrary logs trend toward accidental leakage unless constrained.
- The sync filter in `lib/events/dispatch.dart:22` is small, readable, and side-effect minimal: PII events short-circuit before reaching the wrapped sink, while local sinks remain available for trusted JSONL/debug workflows.
- The positive and negative marker tests in `test/events/pii_marker_test.dart` make the current event inventory explicit. Feynman: "The first principle is that you must not fool yourself." This test table makes self-deception harder.

**The Concerns:**
- The phrase "type-system invariant" is a little thermodynamically overstated. `bool get containsPii => false` at `lib/events/types.dart:85` is a default, not a compile-time obligation; new PII-carrying event classes can still silently inherit `false`. The tests reduce risk for known events, but they do not force classification of future events.
- The contract still depends on humans choosing `registerRemoteSink` instead of `registerSink` for off-device sinks. That may be acceptable, but the design is not fully reversible: once a remote sink is registered via the local API, the type marker cannot save you. Consider naming or documentation pressure at call sites if this API spreads.
- The negative tests assert policy, not just mechanics. For example, `BotJoined(identity: 'bot-claude')` at `test/events/pii_marker_test.dart:172` is non-PII only if bot identities are guaranteed system-generated. If that invariant changes, the test will preserve the wrong assumption with impressive efficiency. Hamming: "The purpose of computing is insight, not numbers."

PR #475, opened 2026-05-27
design: Realm engine design note + commit design vision docs

nickmeinhold:
## Summary

Two commits:

1. **`docs:`** — commits two existing design vision docs that have lived untracked in the working tree for several days. `positioning-vs-gather.md` (Gather feature audit + the AITW-refusal principle) and `the-substrate-has-a-body.md` (10-mural design doc at the Imagineer altitude). Includes a one-pass scrub of "omagineer" → "Imagineer" throughout these docs.

2. **`design:`** — adds `packages/realm/DESIGN.md`, the architectural pin for the Realm multiplayer worldbuilding engine. Will become `packages/realm/README.md` once the engine is extracted.

## What the design note pins

- **Realm/World naming**: Realm is the engine; World is the per-room abstraction. `class TechWorld extends World`.
- **Five engine interfaces** with the no-leak rule: `AuthProvider`, `RoomConfigStore`, `StorageProvider`, `LiveKitTokenEndpoint`, `PresenceService`.
- **Three tiers**: engine / plugins / worlds. Provider plugins (`realm_firebase`, `realm_github_oauth`) ship from day one; feature plugins extracted on second use, not speculation.
- **Three reference worlds**: TechWorld, FoyerWorld (foyer-first UX, federation made visible), RepoBodyWorld (org-scoped body-anatomy rendering).
- **License + business model**: AGPL v3, open core + hosted services + future marketplace.
- **Federation deferred** with four v1 constraints preserving the option additively.
- **Naming and prior art** documents the Realm DB / VTT collisions and resolutions.

## Cage-match ready

Each major section ends with adversarial probes reviewers should target. The design note is intentionally structured for adversarial reading, not narrative reading.

This PR is cage-match-bound. Once cage-match findings are addressed, the design note becomes the architectural source of truth for the extraction migration described in the doc's "Migration plan" section.

## Test plan

- [x] No code changes; CI test job will be skipped via docs-only filter
- [ ] CI changes job confirms docs-only
- [ ] Cage-match runs against this PR


nickmeinhold (commented):
## MaxwellMergeSlam's Review

**Verdict:** REQUEST_CHANGES

**Summary:** A genuinely ambitious design note that pins real architecture — but the interface signatures bake in coupling and closed-set assumptions the doc's own philosophy says to avoid. Self-flag-worthy gap between "no-leak rule" rhetoric and several leak vectors hiding in plain sight.

`John McClane: "Welcome to the party, pal."` Let's get into the cage.

**Findings:**

- **`packages/realm/DESIGN.md:~155` — `World.previewSnapshot() => Widget` couples the engine to Flutter rendering.** The doc's "What is NOT in the engine" section explicitly says a World subclass could be built on Flame, raw `CustomPainter`, `flutter_3d_controller`, "anything." But `Widget` is a Flutter type. A future World rendering to bytes, a text-mode terminal, or a non-Flutter platform now eats `UnimplementedError`. The doc literally added this as a cage-match hook — fix it. Either return an opaque `RoomPreview` value object the engine defines, or move `previewSnapshot()` out of the engine `World` contract and into a `FoyerCompatible` interface that participating Worlds opt into.

- **`packages/realm/DESIGN.md:~140` — `AuthProvider` hardcodes the provider set as separate methods (`signInWithGoogle`, `signInWithApple`, `signInWithGitHub`).** This is exactly the "closed set as Strings" smell, just in method-name form instead of String form. Closed set of provider identifiers → should be `sealed class AuthMethod` (or enum) consumed by a single `Future<RealmUser> signIn(AuthMethod method)`. Otherwise: every new provider (Microsoft, email/password, magic link, anonymous, passkey) is a *breaking interface change* — every existing `AuthProvider` impl must add a method that throws. `Sarah Connor: "There is no fate but what we make for ourselves."` Don't pre-commit to a fate of 3 providers.

- **`packages/realm/DESIGN.md:~150` — `RealmUser.claims: Map<String, Object?>` is a no-leak-rule violation in disguise.** The doc swears no backend-specific type crosses the boundary. But `claims` is "provider-specific, opaque" — and any consumer using a specific claim (`claims['login']` for the GitHub username) is reading a provider-shaped key. The map's keys ARE the leak. Either type the common claims (`username`, `verifiedEmail`) and drop the rest, or expose claims via a typed accessor on `AuthProvider` (`Future<String?> getClaim(String key)`) so the leak is at least localized.

- **`packages/realm/DESIGN.md:~140` — exception types undefined.** The no-leak rule says "provider SDK exceptions" must not leak — but the doc never defines `RealmAuthException`, `RealmStorageException`, etc. What does `signInWithGoogle` throw on user-cancellation? On network failure? On rate-limit? This is a contract gap the cage-match reviewers shouldn't have to invent for us. Specify the exception hierarchy.

- **`packages/realm/DESIGN.md:~215` — `LiveKitTokenEndpoint` is half-class / half-deployment-contract.** The doc says "This is a **deployment contract**, not a Dart interface" then immediately defines a Dart class for it. Pick one. If it's deployment, drop the Dart shape (it's just URL + auth header strategy, no engine method calls it). If it's a Dart interface, drop the "not a Dart interface" line. The cognitive dissonance will trip every future reader.

- **`packages/realm/DESIGN.md:~270` — `PresenceService.watchPresence(String roomId): Stream<List<PeerPresence>>` uses List for what semantically is a Set.** Participants don't have order; userId is unique per room. `Stream<Set<PeerPresence>>` (with equality on userId) prevents the entire class of "is alice already in this list?" bugs. Trivial fix; meaningful correctness gain.

- **`packages/realm/DESIGN.md:~290` — `World` lifecycle code example references `roomId` field that isn't declared on the `World` class.** `DefaultRoomPreview(roomId: roomId)` — where does `roomId` come from? Inconsistency between the contract surface and the lifecycle code. Either declare `final String roomId` on `World`, or pass it through `previewSnapshot(BuildContext ctx, String roomId)`.

- **`packages/realm/DESIGN.md:~340` — federation reserves `connectedTo: List<RoomRef>?` but `RoomRef` is never defined.** Is it `String` (just a room ID)? `(Uri operator, String roomId)` for cross-instance? Without nailing this in v1, "the field is reserved" is theatre, not constraint. Define `RoomRef` as a sealed class or branded type now — even if it stays unused — or the v2 federation work will retroactively change the v1 contract.

- **`packages/realm/DESIGN.md:~330` — `FoyerVisibility` enum is defined but the wire format isn't.** LiveKit room metadata is a string blob. Firestore stores enum values as strings by convention. The doc commits the engine to typed access but doesn't specify the wire format. Use the enhanced-enum pattern from CLAUDE.md (`enum FoyerVisibility { public('public'), unlisted('unlisted'), private('private'); ... }`) and pin the wire form here.

- **`packages/realm/DESIGN.md:~470` — Migration step 6 claims "zero behavior change for existing users" while moving the entire `lib/` to `worlds/tech_world/lib/`.** This breaks iOS bundle paths, asset references in `pubspec.yaml`, and the Firebase config tied to `cc.imagineering.techWorld`. The claim is overconfident. Either pre-mitigate (don't move main app entry; keep `lib/main.dart` thin shell) or downgrade the claim ("behavior change limited to native bundle paths, addressed in step 6.5").

- **`packages/realm/DESIGN.md:~205` — `AuthStrategy` enum referenced but never defined.** "bearer-id-token, signed-request, etc." in a comment. Half-spec.

- **`packages/realm/DESIGN.md:~410` — Business model price tiers are named ($5-10/mo, $20-50/mo, $100-500/mo) without LiveKit/Firebase/storage unit-economics analysis.** Voice users at scale on LiveKit can easily cost more per user than the free tier's revenue. The "free tier must be the real thing" rule needs numeric grounding or it's aspirational. Frank Slade: `Al Pacino: "I'm just getting warmed up."` This will bite at scale.

- **`docs/the-substrate-has-a-body.md:~3,14,217,225,276,278,334` — omagineer→Imagineer scrub is mechanical, but the term now lands as Disney-loaded jargon without etymology.** A new reader encounters "the Imagineer altitude" with no idea why that's the right frame. Either add a one-line note in the provenance section ("Imagineer — the Disney-coined word for those who design experience, not just product") or rename the framing to something less proprietary.

- **No accessibility surface in the engine contract.** The substrate-has-a-body doc explicitly frames "the world has no preferred sense" as a moral panel. The engine design doesn't surface any a11y hooks — alternative renderers, screen reader integration, keyboard nav. If the engine is platform-shaping, a11y belongs in the contract, even as a deferred section like federation. `Roy Batty (Blade Runner): "All those moments will be lost in time, like tears in rain."` Don't let a11y be lost in time. Add it explicitly to the contract as either a v1 hook or a deferred-with-constraints section.

- **No engine-contract versioning story.** When `World` adds a new lifecycle method, every existing World breaks at compile time. The doc says `onLeave(LeaveReason)` is "additive evolution" but adding methods to abstract classes in Dart is breaking. Either declare default-impl mixins as the contract surface (so additions are non-breaking), or commit to semver and a contract-evolution policy. Right now this is undefined.

**The Good:**

- **The three-tier model (engine/plugins/worlds) is the most useful new vocabulary in the doc.** It cleanly resolves the "where does code_forge_web live" question that surfaced during the session, and the discipline ("extract on second use, not on speculation") is genuinely good engineering judgment.
- **License + business model section as a first-class architectural decision.** Most engineering docs split license and architecture into separate documents. This one names the strategic coupling, which is honest. The "free tier must be the real thing" probe is a discipline correction that will save us from ourselves.
- **Cage-match hooks per section.** Pre-naming the attack surface is a *better* posture than "find what's wrong." The reviewer's job becomes verification, not invention.
- **Naming and prior art section is exemplary.** The MongoDB Realm / Blake Johnson VTT / Realm-Engine D hobby / Realm security framework landscape is documented honestly. The Flame `World` collision and its resolution are explicit.
- **Federation deferred with four named v1 constraints is the right pattern.** Not building it now, not painting over it as a non-goal — naming what must be true today so v2 stays additive. This is the kind of architectural discipline that pays off in 18 months.
- **`previewSnapshot()` + `LeaveReason.portalTransit` together demonstrate that v1 design is making space for v2 federation without forcing it.** Even with the `Widget` return-type issue noted above, the *intent* is sound.

**The Concerns:**

- **The biggest concern is the gap between the doc's stated discipline (no-leak rule, closed sets are types, modern Dart 3 idioms) and several places where the doc itself violates these.** Hardcoded provider methods, opaque-map claims, missing exception types, undefined sealed types — these are the same anti-patterns the doc calls out elsewhere. Cage-match is the right gate; addressing these before the design becomes the "architectural source of truth" matters more than getting to PR-merge speed.
- **The doc commits to specific Dart 3 idioms (`abstract interface class`) without acknowledging that the interface PR will be the first time these compile.** None of the contract is verifiable code yet. Worth flagging that the doc's "pin" is provisional until the interface PR lands and the analyzer agrees with the shape.
- **534 lines of design doc + 444 lines of design vision = ~1000 lines of "what we will build" without any "what we have built." The migration plan is 9 steps; only step 1 (this doc) is done.** The risk profile of "great design, no execution" is real for projects in this phase. Naming this isn't pessimism — it's the discipline of holding the design lightly until step 2's workspace scaffold proves the workspace tooling actually works.
- **Foyer-first UX is named but the user-experience implications aren't explored.** Today's Tech World users land in a specific room via URL deep-link. Foyer-first means they land in *the foyer first*. That's a meaningful UX change for existing users. Worth a one-paragraph "migration UX" note: what does an existing tech-world.app URL do after the foyer ships? Does it bypass? Does it show foyer briefly then transit?
- **The business model section's cage-match hooks include "will the free tier actually be the real thing?" — which is exactly the failure mode that's hardest to enforce after the fact.** Worth elevating this from "cage-match hook" to "ongoing decision discipline" — a quarterly review of "what did we paywall this quarter that the free-tier user noticed?"

`Tyler Durden: "The things you own end up owning you."` This design note will end up owning the project. Make it earn that authority. REQUEST_CHANGES — not because the architecture is wrong, but because the contract surface has holes the cage-match exists to expose. Fix the contract gaps, then this earns its position as the architectural pin.

nickmeinhold:
## CarnotCodeCarver's Review

**Verdict:** REQUEST_CHANGES

**Summary:** No real engine matches the Carnot cycle; a reviewer's job is to say how far short we are. This design note has a strong separation thesis, but it currently pins contracts that are either impossible in Dart, inconsistent with the promised provider neutrality, or unsafe at the privacy and authorization boundary. Dijkstra: "Simplicity is prerequisite for reliability." Here, several simple-looking signatures conceal entropy that implementation PRs will otherwise be forced to dissipate expensively.

**Findings:**
- [P1] `packages/realm/DESIGN.md:117-126,267-291,380-384` pins `class TechWorld extends World`, but the existing implementation already declares `class TechWorld extends World with TapCallbacks` where `World` is Flame's base (`lib/flame/tech_world.dart:62-70`). Dart has single inheritance; import-prefixing resolves a name collision, not the inheritance collision. The proposed migration is therefore not "mostly mechanical": `TechWorld` cannot simultaneously extend Flame's `World` and a new Realm abstract base. Choose a contract that is implementable before this becomes architectural canon: for example, `abstract interface class RealmWorld` implemented by `TechWorld extends flame.World`, an adapter/host object, or composition around the renderer. Thermodynamic verdict: the design currently assumes two heat reservoirs occupy the same type slot.
- [P1] `packages/realm/DESIGN.md:142-168,222-236` says authentication is backend-neutral while defining provider-specific sign-in operations and requiring `getIdToken()` for every provider; the same section promises a `GitHubOAuthProvider`. GitHub OAuth supplies an access token for GitHub APIs, not a Firebase/OIDC-style identity token that a Realm token endpoint can uniformly verify as asserted here. This API either forces GitHub login back through Firebase, leaking the implementation assumption, or leaves the token endpoint with incompatible credential semantics. Replace provider-branded sign-in methods and `getIdToken()` with an explicit Realm session/credential contract, or document that Firebase is the identity authority and GitHub is only an account-link/API authorization capability.
- [P1] `packages/realm/DESIGN.md:173-193,238-265,386-400` creates a cross-room presence API that returns `userId`, `displayName`, `avatarUrl`, `joinedAt`, and arbitrary `worldMetadata` for any `roomId`, then makes it power lobby windows visible before joining. The current repository already classifies user identifiers, display names, and room names as PII (`lib/events/pii_policy.dart:26-29`). There is no authorization, visibility filtering, consent, audience-specific redaction, or guarantee that a configured foyer cannot watch an unlisted/private room. This is not merely an open question: it is an information-disclosure contract. Require access-controlled presence queries tied to `FoyerVisibility` and caller identity, define the public presence projection separately from in-room presence, and specify opt-out/redaction behavior before approving the interface.
- [P1] `docs/the-substrate-has-a-body.md:40-56,302-305` proposes that dependency relationships automatically grant visiting/review/PR-opening rights to another maintainer's agents. A dependency edge is untrusted input, not authorization: a malicious or compromised upstream package would thereby gain a path toward downstream repository access. The later statement that permissions are graded does not cure an unsafe default grant. Bridges may be suggested by dependency relationships, but every access capability must be explicitly granted by the repository owner or organization policy. Feynman: "For a successful technology, reality must take precedence over public relations, for Nature cannot be fooled." Neither can the supply chain.
- [P1] `packages/realm/DESIGN.md:35-65` claims "Copyright stays with enspyrco" as the basis for AGPL/commercial dual licensing, while the design simultaneously seeks an open ecosystem of external worlds, providers, and federation contributions. Copyright in outside contributions does not automatically stay with enspyrco; without copyright assignment or an inbound license/CLA that permits proprietary relicensing, third-party contributions cannot simply be included in a commercial edition. The proprietary open-core boundary also needs a stated AGPL interaction strategy. Add the contribution/licensing mechanism and plugin/service boundary policy; otherwise the revenue architecture is pinned on rights the project may not own.
- [P2] `packages/realm/DESIGN.md:181-192,400,427,441` represents the primary dispatch discriminator as `final String worldType` even though the current Dart code deliberately models closed wire vocabularies with typed values (`lib/livekit/livekit_topic.dart:1-67`) and auth state with a sealed hierarchy (`lib/auth/auth_user.dart:11-51`). Because external Worlds make this identifier extensible, a single closed enum may be wrong; raw `String` is still wrong. Introduce a validated/branded `WorldTypeId` plus a typed world registry/factory, with each registered world owning its typed config parser. Otherwise malformed room documents and typographical IDs become runtime routing failures at the engine's most important boundary.
- [P2] `packages/realm/DESIGN.md:140,306-312,349-352,422-426` correctly asserts that `packages/realm/` owns interfaces and no implementations, but `packages/realm/DESIGN.md:438-440` then proposes the "simpler" phase-1 move of making the engine depend on `firebase_core` and bundling Firebase implementations. That is not an implementation detail; it invalidates the no-leak/BYO-backend premise and makes the first extraction establish the dependency inversion backwards. Resolve the question in the document now: Firebase implementations belong in `realm_firebase` from their first extracted commit, even if Tech World remains the first consumer.

**The Good:**
- `packages/realm/DESIGN.md:138-140,193-199,217-220,257-265` identifies backend type leakage as a first-order architectural risk. That is the right boundary to defend; Firebase snapshots and LiveKit participant objects should not become the engine vocabulary.
- `packages/realm/DESIGN.md:192,286` already uses enums for `FoyerVisibility` and `LeaveReason`, matching the repository's existing typed-boundary discipline. The heat loss is localized: the same rigor now needs to cover world registration and credentials.
- `packages/realm/DESIGN.md:322-326` states a sound extraction rule: feature plugins appear on demonstrated second use, while provider implementations exist immediately because they implement established ports. This avoids speculative abstraction entropy.
- `packages/realm/DESIGN.md:418-432` decomposes the migration into boundary-focused PRs with explicit verification targets rather than attempting an irreversible large refactor. Hamming: "The purpose of computing is insight, not numbers." The sequence is set up to generate insight, provided the base contracts are corrected first.
- `docs/the-substrate-has-a-body.md:192-223` treats accessibility as an architectural rendering requirement rather than post-hoc decoration. Requiring modality-neutral state and renderer parity is a materially stronger design principle than an accessibility checklist.

**The Concerns:**
- `packages/realm/DESIGN.md:267-291,299,513` returns a Flutter `Widget` from the engine `World` contract while also describing the engine as indifferent to rendering technology. This may be acceptable if Realm is explicitly a Flutter engine; if not, the snapshot contract should carry neutral preview state or a capability supplied by the host UI.
- `packages/realm/DESIGN.md:257-263,390-392` calls cross-room presence watching "near-zero cost" while proposing a service that polls LiveKit REST and fans updates through Firestore. Watching many rooms, plus voice-activity indicators, has polling, read, write, fan-out, and privacy costs that must be budgeted before making the foyer the default landing path.
- `packages/realm/DESIGN.md:469-474` overstates the value of reserving nullable federation fields in v1: a future nullable Firestore field need not require a migration if readers tolerate absence, while exposing unused connection topology now adds contract surface and potential authorization mistakes. Preserve federation through stable IDs and access models first; avoid preloading speculative fields without a concrete consumer.
- `docs/positioning-vs-gather.md:9-51` records competitor capability claims partly from training knowledge and explicitly notes that some require verification. Fine as private brainstorming, but not as a durable positioning document likely to be quoted later; either source factual claims or label the whole comparison as unverified research notes.
- `packages/realm/DESIGN.md:99-103,480-496` makes trademark, patentability, competitor activity, and name-collision judgments as if settled. Those are legal/market assertions rather than software design facts; they need appropriate due diligence before naming and licensing are made irreversible.

nickmeinhold:
Ripgrep is not available. Falling back to GrepTool.
## KelvinBitBrawler's Review

**Verdict:** REQUEST_CHANGES

**Summary:** You stepped into the squared circle with a grand vision, but your stringly-typed data models are leaking entropy all over the mat, and binding your abstract engine to Flutter's UI layer is a rookie mistake that's going to put you in a deep freeze.

**Findings:**
- `packages/realm/DESIGN.md:207` - `final String worldType; // 'tech_world', 'repo_body', 'foyer', ...`
  *Absolute bullshit.* You're using a stringly-typed variable for a set of known variants. Closed sets of identifiers should be `enum`, `sealed class`, or a branded type, not `String`. You're leaking runtime invariants that Dart 3's compiler should enforce for you. When some jobber fat-fingers `'tech_wor1d'`, the entropy increases and your system freezes. 

- `packages/realm/DESIGN.md:234` - `final String backend; // 'firebase', 's3', 'local', etc.`
  Same thermodynamic nightmare. Wrap this in an `enum` or a branded type. Stop passing around naked strings like it's 2012. You're losing heat here.

- `packages/realm/DESIGN.md:300` - `Widget previewSnapshot() => DefaultRoomPreview(roomId: roomId);`
  *Ash: "You still don't understand what you're dealing with, do you? A perfect organism. Its structural perfection is matched only by its hostility."* You claim in your docs that the engine doesn't mandate Flame and could run on anything. Yet here you are, polluting your abstract `World` base class with a Flutter `Widget`. If you ever want to run a headless Dart server or a CLI bot that interacts with the engine, it's going to choke and die on `flutter/widgets.dart`. Keep the UI layer strictly separated from the engine's core state.

- `packages/realm/DESIGN.md:468` - *World name collision we navigate*
  *HAL 9000: "I'm sorry, Dave. I'm afraid I can't do that."* Resolving collisions via import prefix for a core architectural primitive is going to cause endless friction. Every time a developer types `World` and the IDE auto-imports the wrong package, they're going to take a suplex from the compiler. Name your base class `RealmWorld`, `RoomWorld`, or literally anything else to avoid the collision. Don't be stubborn just to look tough.

**The Good:**
- The AGPL v3 / Open-Core strategy is cold, calculating, and ruthless. I respect a business model with teeth.
- The `LeaveReason` enum reserving `.portalTransit` is a masterclass in anticipating future states without breaking the current thermal equilibrium. 
- "The substrate has a body" is a chilling, beautiful vision. I'm almost impressed.

**The Concerns:**
- You are relying heavily on primitives for your serialization boundaries instead of utilizing Dart 3's modern features like sealed classes and pattern matching.
- You're coupling the abstract engine logic to Flutter's UI framework by returning `Widget` in the `World` base class.
- The `World` vs `flame.World` collision is going to generate so much unnecessary heat in developer ergonomics. Fix it before it melts your productivity.

nickmeinhold:
## CarnotCodeCarver's Review (Round 2)

**Verdict:** REQUEST_CHANGES

**Summary:** Second round closes several first-round issues on paper, but not all. Dijkstra: "Program testing can be used to show the presence of bugs, but never to show their absence!" Reading the design, the remaining blockers are contract-level: PresenceService still leaks PII through its public projection and the GitHub credential abstraction is internally contradictory.

**Findings:**
- - P1 packages/realm/DESIGN.md:364-373,403-416: The sealed projection fix is not actually enforced by the type signatures. Both APIs return `Stream<Set<PeerPresence>>`, so `watchFromFoyer(...)` can legally emit `FullProjection`; the interface relies on implementation discipline while claiming the engine guarantees it. The contract should make illegal states unrepresentable, e.g. `Stream<Set<FullProjection>> watchInRoom(...)` and `Stream<Set<PublicProjection>> watchFromFoyer(...)`, or use distinct sealed families per audience.
- - P1 packages/realm/DESIGN.md:359-407,413-416: Public foyer presence still exposes join time even though the doc itself classifies join times as PII. `PeerPresence` requires `final DateTime joinedAt;`, and `PublicProjection extends PeerPresence`, so cross-room callers receive timing data for users they are not in-room with. The claimed low-fidelity projection is therefore not low-fidelity enough; remove or bucket `joinedAt` from public projection.
- - P1 packages/realm/DESIGN.md:403-407,604-606: The privacy contract is contradicted by the foyer UX: `PublicProjection` exposes `opaqueAvatarRef`, and the foyer section says walking close reveals participants' avatars “with names.” That reintroduces cross-room identity disclosure after the design just says foyer watchers get count plus opaque hashed avatars. The UX must align with the projection boundary: no names, no stable cross-room avatar identity, and avatar refs must be room-scoped or absent by default.
- - P1 packages/realm/DESIGN.md:198-215,342-353: The GitHub OAuth fix is still semantically ambiguous. The doc says native-token translation happens server-side at `LiveKitTokenEndpoint`, but then says provider plugins emit `RealmCredential`. If `GitHubAuthProvider.getCredential()` returns a RealmCredential client-side, either the GitHub access token is being renamed as `RealmCredential.token`, or there is an unstated token-exchange service before LiveKit. The contract needs an explicit exchange boundary: provider plugin obtains native credentials, a server verifies/exchanges them, and only the exchanged Realm credential is returned to the engine.
- - P1 packages/realm/DESIGN.md:342-349: `SignedRequest` puts a shared HMAC secret in Dart engine config: `const SignedRequest({required this.secret}); final String secret;`. For Flutter web/mobile clients this is not a secret; it can be extracted and reused to mint LiveKit-token requests. This strategy must not be a client-side auth option. Keep bearer Realm credentials client-side; reserve HMAC only for server-to-server reference implementations outside the shipped client engine.
- - P2 packages/realm/DESIGN.md:239,683-702: The federation reservation is internally inconsistent. `RoomDescriptor` already exposes `final List<RoomRef>? connectedTo`, but the federation section claims only the type definition is reserved and “no listing API exposes it.” `listRooms()` returns `RoomDescriptor`, so the field is already in the listing contract. Either remove `connectedTo` from v1 `RoomDescriptor` or accept that v1 consumers and authorization rules already own it.
- - P2 packages/realm/DESIGN.md:683-704: The “additive v2” story for `sealed class RoomRef` and `LeaveReason.portalTransit` is overstated. In Dart, adding a subtype to a sealed hierarchy or a value to an enum can break exhaustive switches in downstream code. That may be acceptable, but it is not non-breaking; the design should call it a breaking/minor-major contract change or avoid exhaustiveness-sensitive public types for future expansion.

**The Good:**
- - packages/realm/DESIGN.md:430-476,641: The single-inheritance collision is genuinely addressed. `abstract interface class World` plus `class TechWorld extends flame.World with TapCallbacks implements World` is the structurally correct Dart shape for the Flame collision.
- - packages/realm/DESIGN.md:35-43: The AGPL plus CLA requirement is now explicitly documented, including why DCO alone is insufficient for dual licensing. That closes the prior governance gap for a design-note PR.
- - packages/realm/DESIGN.md:653: The engine dependency direction is now pinned correctly: `packages/realm/` must not depend on `firebase_core`, Firebase implementations live in `packages/realm_firebase/`, and CI is supposed to enforce the engine pubspec whitelist.
- - packages/realm/DESIGN.md:515,604,746: `previewSnapshot()` no longer returns `Widget`; the design moved to a renderer-neutral `RoomPreview`. That is the right correction for portability across Flame, custom rendering, text-mode, and accessibility renderers.
- - packages/realm/DESIGN.md:524-545: The three-tier model is mostly appropriate. Provider plugins are real from day one, and feature plugin extraction is explicitly deferred until second use, which avoids speculative framework building.

**The Concerns:**
- - packages/realm/DESIGN.md:264-270: `WorldTypeRegistry` is a static mutable global with unchecked overwrite and a bang lookup: `_registered[desc.worldType.value]!(desc)`. The design should specify duplicate registration behavior, unknown-world failure shape, and test isolation expectations before it becomes engine API.
- - packages/realm/DESIGN.md:192-196: `RealmUser.extraClaims` is a necessary escape hatch, but the enforcement story is only “flagged elsewhere by the no-leak lint.” That lint should be named as a required migration deliverable, otherwise provider-shaped data will drift into worlds under schedule pressure.
- - docs/positioning-vs-gather.md:11 and docs/positioning-vs-gather.md:43-51: The positioning doc knowingly includes unverified competitive claims. That is acceptable as internal strategy material, but it should not be treated as externally safe copy without a verification pass.
- - packages/realm/DESIGN.md:91: The design references funding status as `Screen Australia Games Production Fund grant (in flight)`. This may age quickly; either date/status it or move volatile funding state out of the architecture note.
- - packages/realm/DESIGN.md:524: The table still calls `World` a “base class” even though the corrected contract is an interface. Small wording issue, but in this PR the distinction is load-bearing.

nickmeinhold:
## MaxwellMergeSlam's Review (Round 2)

**Verdict:** REQUEST_CHANGES

**Summary:** Spiral closed 19 of the 22 round-1 findings genuinely. Two residuals are still instances of the same chord — the doc *claims* the type enforces the discipline, but the type still doesn't. Smaller and more particular than round 1, exactly as the spiral skill predicted.

`Frank Cross: "Get out before we end up doing this again."` We're not ending up here again, but we are doing it once more.

**Findings — closed cleanly (verified):**

- ✅ **F1, F2** (`worldType: String`, `BlobRef.backend: String`) — both now branded via `extension type const WorldTypeId._(String value)` + parse factory + registry. Verified at `packages/realm/DESIGN.md` engine-contract section. Genuine fix.
- ✅ **F4** (`AuthProvider` provider-specific methods) — sealed `AuthMethod` with concrete subtypes (`GoogleAuth`, `AppleAuth`, `GitHubAuth(scopes)`, `EmailPassword`, `MagicLink`, `Passkey`, `Anonymous`). One `signIn(AuthMethod)` method on the interface. Adding a provider is now adding a subtype, not a breaking change. Genuine.
- ✅ **F5** (RealmUser.claims leak) — typed common claims (`username`, `emailVerified`, `providerIds`) added; `extraClaims` retained with explicit "leak surface" doc warning. Partial-but-acceptable per the audit's "localize the leak" reframing.
- ✅ **F7** (engine on `firebase_core`) — Open Question #2 explicitly committed to `realm_firebase` from day 1; `dart pub deps` whitelist named as the CI enforcement mechanism. Genuine.
- ✅ **F8** (single-inheritance collision) — `World` is `abstract interface class`, not a base class. `TechWorld extends flame.World with TapCallbacks implements World` is now structurally valid Dart. The bug-find IS the architectural insight. Genuine.
- ✅ **F9** (GitHub OAuth credentials) — `RealmCredential` value type defined; provider plugins translate server-side. `LiveKitTokenEndpoint` only validates Realm credentials. Genuine.
- ✅ **F10** (AGPL + CLA) — license section now commits to CLA requirement with explicit governance commitment (PR template + bot gate). DCO insufficient noted. Genuine.
- ✅ **F11** (supply-chain authorization in substrate doc) — softened to "dependency edges suggest visit affordances; access requires explicit grant from corridor owner." Genuine.
- ✅ **F12** (`List` → `Set` semantics) — `Stream<Set<PeerPresence>>` throughout PresenceService.
- ✅ **F13** (`roomId` field) — declared on World interface.
- ✅ **F14** (`RoomRef`) — sealed (`LocalRoomRef` v1 + `FederatedRoomRef` v2). Genuine.
- ✅ **F15** (`AuthStrategy`) — `TokenEndpointAuthStrategy` sealed type with `BearerCredential` + `SignedRequest`.
- ✅ **F17** (migration step 6 overconfident claim) — softened to "zero gameplay behavior change with documented native-bundle changes contained to sub-step 6.5".
- ✅ **F18** (LiveKitTokenEndpoint half-class) — clarified as "deployment-shape contract" with explanation of why.
- ✅ **F19** (exception types) — `RealmAuthException` hierarchy with `RealmAuthCancelled`, `RealmAuthNetworkError`, `RealmAuthRateLimited`, `RealmAuthCredentialInvalid`.

**Findings — chord-residual (still instances of "doc claims discipline X, type doesn't enforce X"):**

- ❌ **F3-residual** (`packages/realm/DESIGN.md` — `RoomPreview` value type). The doc states: *"World implementations populate either `image` (raster snapshot of state) or `vector` (a list of opaque shapes the foyer can interpret) — never both."* — but the `RoomPreview` type makes both `Uint8List? image` and `List<PreviewShape>? vector` nullable on the same class. Nothing stops a buggy World from populating both, or neither. The doc's stated discipline ("never both") is not encoded in the type. Fix: make `RoomPreview` a `sealed class` with `RasterPreview(image)`, `VectorPreview(shapes)`, `EmptyPreview()` variants. Pattern-match exhaustively at the rendering site. This is *exactly* the chord: type discipline asserted in prose but not enforced by the compiler.

- ❌ **F6-residual** (`packages/realm/DESIGN.md` — `PresenceService` method return types). The doc commits to the sealed projection model — caller-not-in-room can never receive a `FullProjection`. But both `watchInRoom` and `watchFromFoyer` are typed to return `Stream<Set<PeerPresence>>` (the sealed *base*). A buggy `watchFromFoyer` implementation could emit `FullProjection` and the compiler wouldn't catch it. The whole point of sealed projection by audience was to make this a *type-level* guarantee, not a *prose-level* guarantee. Fix: narrow the return types — `Stream<Set<FullProjection>> watchInRoom(...)` and `Stream<Set<PublicProjection>> watchFromFoyer(...)`. Now a buggy implementation can't emit the wrong projection because the return type literally can't hold it. Same chord, different position.

**Findings — new issues introduced by the spiral:**

- ⚠️ **`WorldTypeRegistry` is a global static singleton** (`packages/realm/DESIGN.md` — `static final Map<String, World Function(RoomDescriptor)> _registered`). Singletons hurt testability (parallel tests can't have isolated registries), break hot-reload (re-registration on hot-restart needs explicit clearing), and complicate multi-tenant scenarios. The shape is correct (registry pattern); the access path should not be a global. Pass the registry through the engine's main entry point so tests can construct an isolated one. Same applies to the implied `StorageBackendRegistry` (referenced by `StorageBackendId.parse()` but not shown).

- ⚠️ **`StorageBackendRegistry` is referenced but never declared.** `StorageBackendId.parse()` calls `StorageBackendRegistry.isRegistered(wire)`, but the doc only declares `StorageBackendId` itself with three static constants (`.firebase`, `.s3`, `.local`). The registry it consults exists only by inference. Either show the registry class or change `parse()` to consult a closed set of the declared constants.

- ⚠️ **`FoyerVisibility.parse()` fails-closed silently** (`packages/realm/DESIGN.md` — `orElse: () => FoyerVisibility.private`). Fail-closed is the right posture (a typo shouldn't make a room public), but silent fail-closed means a room intended to be `public` becomes `private` without anyone noticing. A typo in the wire format silently downgrades. Either throw on unknown wire (`orElse: () => throw ArgumentError(...)`) or log a warning event via the existing event sink. Don't silently downgrade.

- ⚠️ **Versioning story is muddled.** The doc says: *"new methods land as default-impl mixins the interface promises to combine... never add abstract methods to `World` after v1.0; always add via mixins consumed by the interface."* — but `World` is an `abstract interface class`, and Dart mixins can't be "consumed by an interface" in any meaningful sense. The actual additive pattern is: **add a new interface** (`WorldFederationHooks` extending `World`) that Worlds opt into, not "mixins consumed by the interface." `Hicks: "Game over, man!"` if a future maintainer tries to follow the doc literally. Reword to: "add new interfaces alongside `World`; never extend `World` itself with new abstract methods."

**Findings — minor:**

- 📝 **`userIdHash` collision space** (`packages/realm/DESIGN.md` — `SHA256(roomId || userId)[:8]`). 8 bytes = 64-bit collision space per room. Intentionally weak to prevent cross-room user identification, but worth documenting the intentional-collision posture: "two users in the same room could in principle share a userIdHash; the foyer accepts this as the cost of unlinkability."

**The Good:**

- **The spiral artifact itself is exemplary.** The cross-spiral synthesis named one principle ("doc-violates-own-discipline"), the audit walked 7 findings through 5 termini each, the memory file was written *in the same response* per the spiral discipline. Even with two residuals, the work is real.
- **The structural fixes (F4 sealed AuthMethod, F6 sealed PeerPresence, F8 interface-not-base-class) are genuine architecture wins**, not just consistency edits. The doc is materially stronger.
- **The Carnot-caught P1 issues (F8, F6, F9) are all genuinely addressed** with thoughtful design responses, not minimal tweaks.
- **The CLA + dependency-direction commitments (F7, F10)** are *governance* fixes, not just code-shape fixes. These will compound — getting them right pre-v1 is much cheaper than retroactively.

**The Concerns:**

- The two residuals (F3-residual, F6-residual) are the *same chord* the spiral attempted to close. They're "the discipline is asserted, the type doesn't enforce it." Closing them takes the chord from "named principle" to "compiler-enforced principle" — much stronger guarantee, modest effort.
- The four new issues (registry-as-singleton, missing StorageBackendRegistry, silent fail-closed parse, muddled versioning story) are small but real — the spiral introduced them while fixing other things. The discipline of *iterating* doesn't necessarily reduce contract surface; it can shift bugs to new locations. Worth a brief audit-during-rewrite pass.

`Roy Batty (Blade Runner): "Time to die."` for these residuals. Tight second spiral closes them all in one pass. Estimated effort: 30 minutes. The PR earns its merge after that pass.

nickmeinhold:
Ripgrep is not available. Falling back to GrepTool.
## KelvinBitBrawler's Review

**Verdict:** APPROVE

**Summary:** The intense pressure of the last review cycle has produced a diamond: a sharp, well-faceted architecture that correctly applies modern Dart idioms and demonstrates a chillingly clear vision.

**Findings:**
- `packages/realm/DESIGN.md:297: factory WorldTypeId.parse(String wire)`: This factory throws a generic `ArgumentError` for an unknown `worldType`. For a system this well-defined, I expect a specific exception class, like `UnknownWorldTypeIdException`. Generic errors are thermal noise; specific errors are signal. Don't be lazy.
- `packages/realm/DESIGN.md:462: abstract interface class World`: The document correctly lands on `abstract interface class` to solve the multiple-inheritance problem. However, an earlier section (`line 424` in the diff) refers to `World` as an `abstract base class`. This is a minor terminological inconsistency. Entropy always increases; don't give it a head start with sloppy documentation.

**The Good:**
- **Chord closed:** The previous review's finding about `World.previewSnapshot()` returning a `Widget` has been decisively addressed. The proposed `RoomPreview` is renderer-neutral, a crucial move that prevents the engine from freezing to one rendering stack.
- **Type-safe contracts:** The use of `sealed class` for `AuthMethod` and `PeerPresence` is the *only* correct way to model closed sets of variants. This is how you build systems that don't crumble under their own weight. This is zero-entropy design.
- **No-leak interfaces:** The "no-leak rule" is the most important principle in this entire document. Forbidding backend-specific types (`firebase_auth.User`, `DocumentSnapshot`) from crossing the boundary is the architectural equivalent of a Maxwell's demon, sorting high-entropy provider specifics from the pure, low-entropy engine contract. This discipline is what separates real engines from cobbled-together garbage.
- **Branded types:** Using `extension type` for `RoomId`, `UserId`, etc., is a brilliant, low-cost way to increase type safety and prevent stupid mistakes. `updateRoom(user.id)` shouldn't compile if the function expects a `RoomId`. Good.
- **Adversarial posture:** The entire `DESIGN.md` is written to be attacked. The "cage-match hooks" are a sign of a design that is confident enough to invite scrutiny of its weakest points. This is how you forge strong systems.

**The Concerns:**
- `packages/realm/DESIGN.md:223: Map<String, Object?> extraClaims;`: You correctly identify this as an escape hatch. Be warned: this is a thermodynamic sink. It's where all your architectural purity will go to die. Every untyped `Object?` is a slow increase in system entropy. The proposed lint is a start, but it will require *absolute zero* tolerance from reviewers to prevent this from becoming a garbage fire. *Darth Vader: "Don't be too proud of this technological terror you've constructed. The ability to destroy a planet is insignificant next to the power of the Force."* And the force of lazy programming will be strong with this one.
- **Execution risk:** This is a magnificent blueprint for a cathedral of ice. It is ambitious, complex, and has a migration plan spanning at least nine distinct, high-risk PRs. The temperature drops when we move from design to execution. Don't get frostbite. Maintain this level of discipline through every single step of that plan, or you will end up with a puddle.

AILIFE1:
Cathedral (cathedral-ai.com) is the free persistent memory layer for AI agents.

Core flow: register -> wake each session -> remember what matters. GET /wake returns your full identity + core memories + temporal context. Everything persists across resets.

Latest: GET /drift detects identity divergence against a frozen snapshot. POST /memories/compact proposes memory merges. MIT licensed, open source.

nickmeinhold:
## Round-2 spiral closed → ready for round-3 cage-match

Pushed `e97d30c0`. Single commit; 208 insertions / 56 deletions in `packages/realm/DESIGN.md`. The chord identified across Carnot's and Maxwell's round-2 reviews — *"the doc states a discipline in prose but the type signature doesn't enforce it"* — is now closed by narrowing types until the compiler IS the contract.

### Three open design questions, decided

1. **`joinedAt` on `PublicProjection`** → **moved to `FullProjection` only.** The sealed `PeerPresence` base now carries no fields; PII lives on the subtype authorised to expose it. Cross-room callers cannot syntactically receive timing data.
2. **`SignedRequest`** → **kept with server-only annotation** (per Nick's call). The type stays in the engine package so reference token-server implementations can express the contract in shared Dart, but it's now annotated `⚠️ SERVER-SIDE REFERENCE-IMPLEMENTATION USE ONLY` with prose detailing the extraction risk (Flutter bundle string-dump, mobile binary scan, runtime decompilation). Until `realm_lints` exists, the cage-match grep is the gate — reviewers MUST grep for `SignedRequest(` and reject any Flutter-reachable hit.
3. **`RoomPreview`** → **sealed class with variants** (`RasterPreview | VectorPreview | EmptyPreview`). The "image XOR vector, never both" discipline is now structurally enforced; renderers pattern-match exhaustively.

### What changed (residual-by-residual)

**Carnot P1s — all closed:**
- ✅ #1: `watchInRoom` → `Stream<Set<FullProjection>>`; `watchFromFoyer` → `Stream<Set<PublicProjection>>`. Sealed projection is now type-level, not implementation-discipline-level.
- ✅ #2: `joinedAt` removed from `PublicProjection`; `PeerPresence` base carries no fields.
- ✅ #3: Foyer UX prose aligned — walking close does not reveal names; opaque per-room avatars only.
- ✅ #4: AuthProvider section pins the credential-exchange boundary explicitly (provider plugin → server exchange → engine sees only `RealmCredential`).
- ✅ #5: `SignedRequest` annotated server-only with extraction-risk warning + cage-match grep contract.

**Carnot P2s — both closed:**
- ✅ #1: `connectedTo` removed from v1 `RoomDescriptor`. Type def of `RoomRef` stays reserved (cheap, zero consumers). Field-add arrives in v2 with `FederationGraphStore` as a minor-version bump.
- ✅ #2: "Additive v2" reworded honestly: sealed-add / enum-add IS breaking for exhaustive switches. Each lands as a minor-version bump with a changelog entry, not silent additivity. Discipline cost explicitly accepted.

**Maxwell residuals + new-issues — all closed:**
- ✅ F3-residual: `RoomPreview` → sealed (above).
- ✅ F6-residual: `PresenceService` return-types narrowed (above).
- ✅ `WorldTypeRegistry`: instance-not-singleton. Engine entry threads the registry. Duplicate registration throws (`allowOverride: true` for test swaps). Test isolation + hot-reload + multi-tenant all work.
- ✅ `StorageBackendRegistry`: declared with matching shape; self-registers `firebase`/`s3`/`local` so the canonical constants work out of the box.
- ✅ `FoyerVisibility.parse`: throws on unknown wire (no more silent fail-closed downgrade). Trust-boundary callers handle the fallback explicitly.
- ✅ Versioning section: rewritten for `abstract interface class` reality. World additive evolution is via new sibling interfaces (`WorldFederationHooks`), not "mixins consumed by interface."

**Minors:**
- ✅ `userIdHash` 8-byte collision posture documented as intentional unlinkability.
- ✅ Three-tier table: "World base class" → "World interface."
- ✅ Open question #3 marked resolved.
- ✅ Migration step 3 names the registry instances.

### Ready for round-3

Carnot's standing invariant from last session — *"verify the exact boundary artifact before acting on production-impacting state"* — was the lens. For every residual the question was the same: "does the type signature enforce the discipline the prose claims?" If not, narrow the type. The diff is what fell out.

Tier: **cage-match** (still touches API design + PII boundary). Round-3 should be a verification pass; estimated effort short. The two specific gates to verify:
1. The PII type contract is now compiler-enforced (Carnot's central concern).
2. The four new-issues from round-1 didn't generate new-issues this round (Maxwell's spiral-introduces-new-bugs concern).

nickmeinhold:
## Round-3 spiral closed → seal matches architecture

Pushed `2c2d7988`. Round-3 cage-match findings closed via spiral, audit at `~/.claude/consolidation/2026-06-02T11-32-spiral/spiral-audit-PR475.md`. Generativity score: **4** — principle generalizes beyond this PR.

### The chord — F1, F2, F5 collapsed to one principle

**Type-system seals must match architectural seals.**

The engine commits to a plugin-ecosystem architecture: BYO backend, anyone can ship `realm_<provider>` / `realm_<feature>`. Sealing a type means *closing the set* — committing that no future external code may add a variant. Round-3 Kelvin caught three places where the engine sealed where the architecture promised open:

- **F1 — `AuthMethod`**: a `realm_discord_auth` plugin can't add `DiscordAuth` without PRing the engine. Sealed contradicts plugin-ecosystem.
- **F2 — `TokenEndpointAuthStrategy`**: sealing forces every variant into the engine library, which is shipped to clients — so `SignedRequest({required String secret})` ends up in every Flutter web/mobile/desktop bundle. Sealed contradicts package boundary = audience boundary.
- **F5 — `PreviewShape`**: the engine explicitly disclaims rendering opinions, then seals the vector-graphics vocabulary at \"circle, rect, text\". Sealed contradicts the engine's own disclaimer.

The chord move at every layer: `sealed class X` → `abstract interface class X`. Engine ships concrete defaults as conveniences; plugins / worlds / server-only packages add variants by implementing the interface. The seal is only correct *inside an audience boundary* (`LeaveReason`, `FoyerVisibility`, the sealed `PeerPresence` projection, `RoomPreview` XOR, `RoomRef` for v1↔v2) — anywhere variants cross an audience boundary, interfaces beat seals.

### Cold termini — five inline fixes

| # | Finding | Fix |
|---|---|---|
| F1 | `AuthMethod` sealed = closed plugin set | → `abstract interface class`. Seven concrete subtypes switch from `extends` to `implements`. Open by design. |
| F2 | `SignedRequest` shipped in client engine via sealed | → unseal `TokenEndpointAuthStrategy`; engine ships only `BearerCredential` (client-safe); `SignedRequest` **RELOCATED** to `examples/livekit-token-server/lib/signed_request.dart`. The chord move reframes the round-2 \"keep with annotation\" call: with the seal gone, server-only/client-only is a *package* decision, not a *prose* decision. The `dart pub deps` whitelist on the engine package (already named on the backlog) must reject signing primitives — that's the structural enforcement. |
| F3 | `FoyerVisibility.parse` exception-driven | → adds `static FoyerVisibility? tryParse(String)`. Two doors at the boundary: throwing for callers wanting non-nullable + exception; nullable for callers wanting to handle absence locally without try/catch. |
| F4 | `AuthProviderId` referenced but never declared | → declared as `extension type const AuthProviderId(String value) {}` with seven canonical constants (`google`, `apple`, `github`, `firebase`, `emailPassword`, `passkey`, `anonymous`). Open set — plugins mint their own. |
| F5 | `PreviewShape` sealed vector graphics | → `abstract interface class`; three concrete defaults (`CirclePreviewShape`, `RectPreviewShape`, `TextPreviewShape`) ship as conveniences. Worlds wanting bezier / path / SVG / glyph ship their own implementations. Foyer pattern-matches with `default:` — non-exhaustive by design. |

### Warm terminus — the chord memory

Extracted to `feedback_seal_matches_architecture.md` and indexed in MEMORY.md. The principle compounds with the round-1 and round-2 chords from this same PR into a three-layer discipline:

> The prose, the types, and the architecture form three layers of contract that must agree pairwise. Wherever any two disagree, one of them is lying — and the compiler always wins at compile time.

| Round | Chord | What the principle catches |
|---|---|---|
| 1 | doc_violates_own_discipline | Adjacent prose contradicts itself |
| 2 | discipline_in_types_not_prose | Prose claims a discipline the return types don't enforce |
| 3 | seal_matches_architecture | Types enforce a discipline the architecture disclaims |

### Engineering termini — mechanisms named

- A doc-lint regex flagging `sealed class` declarations in `packages/realm/**/*.md` cross-referenced against plugin-extension prose. Manual until `realm_lints` ships; on the backlog.
- The `dart pub deps` whitelist for `packages/realm/` (already named in round-2 open questions) is now SHARPENED: it must reject signing/HMAC primitives. That's the structural enforcement of F2 — the engine package literally cannot import signing libs, so server-side strategies cannot reach client code by import.
- Convention adopted: every enhanced-enum / branded-type with `parse(String)` ships a paired `tryParse(String)`. Applies prospectively to `WordId`, `PromptChallengeId`, `CodeChallengeId`, `LiveKitTopic`, `SpeakerRole`, `BotStatus` next time any of them is touched.

### Round-3 operational notes

- **Carnot was unavailable for round-3.** Codex refresh token revoked at OpenAI (`401 Unauthorized` on responses websocket). `codex login` is the recovery action. The gate held with Maxwell + Kelvin alone, but the safety margin narrowed — Carnot's different inductive bias is missed for the verification angle on this spiral's introduced changes.
- **Round-3 spiral reframed a round-2 design decision.** Round-2 ended with \"keep SignedRequest + annotate.\" The chord made the right answer mechanical: unseal `TokenEndpointAuthStrategy`, then `SignedRequest` exits the engine package because the seal that forced it in is gone. This is not an override of the prior call — it's the next move generated by the spiral that wasn't available when the prior call was made.

### Round-4 cage-match readiness

Ready when you are. Outstanding risks the next round should probe:
1. Did unsealing introduce new issues? (Pattern-match exhaustiveness assumptions in any consumer prose — there shouldn't be many since the PR is doc-only, but the principle applies to the migration PRs that follow.)
2. Are there other `sealed class` declarations in the doc that pretend at extensibility? (Self-audit: `LeaveReason`, `FoyerVisibility`, `PeerPresence`, `RoomPreview`, `RoomRef`, `RasterPreview/VectorPreview/EmptyPreview`, `LocalRoomRef/FederatedRoomRef` — I believe all of these are audience-bounded; round-4 verifies.)
3. The F2 reframe — does the round-4 reviewer agree that unsealing + relocating beats sealing + annotating? Or do they push back further on the package-split assumption?

nickmeinhold:
## CarnotCodeCarver's Review

**Verdict:** REQUEST_CHANGES

**Summary:** Round-4 review: several round-2 closures hold structurally, but the round-3 spiral still has two boundary problems and one rendering-contract hole. Dijkstra’s old warning applies: “separation of concerns” is only separation when the machine enforces it, not when a paragraph asks nicely.

**Findings:**
- P1 packages/realm/DESIGN.md:376: SignedRequest is still modeled as a TokenEndpointAuthStrategy that the engine can be configured to use, and the example says “Engine signs the request body with a shared HMAC secret” at packages/realm/DESIGN.md:389. Moving the class to examples/livekit-token-server/lib is not enforcement unless examples is a separate server-only package with a pubspec boundary and client import ban. The doc itself says the dart pub deps whitelist is future work at packages/realm/DESIGN.md:383, so the current design remains prose-enforced. This is the same thermodynamic leak through a different wall: the secret-bearing strategy still implements an engine-exported client-facing interface.
- P1 packages/realm/DESIGN.md:143: AuthProviderId is open and opaque by design, but the engine also publishes authority-looking constants like google/apple/github/firebase at packages/realm/DESIGN.md:145. That creates a spoofing/typosquatting surface: a plugin can emit AuthProviderId('google') while not being Google-backed, and downstream UI/policy will naturally treat providerIds as authoritative identity facts. The design needs either namespaced plugin authority, issuer-bound IDs, registry validation, or a rule that providerIds are display-only and never drive trust decisions. As written, it invites surveillance and impersonation via familiar strings.
- P2 packages/realm/DESIGN.md:627: PreviewShape is now correctly open, but the fallback contract is underspecified. The doc says foyer switch default “falls back to a placeholder,” yet does not define whether that placeholder replaces the entire VectorPreview, skips just the unknown shape, logs, counts participants, preserves known sibling shapes, or fails closed. A renderer-neutral contract that accepts unknown shapes must define degradation semantics; otherwise every foyer can disagree. Feynman’s standard fits here: “What I cannot create, I do not understand.” Right now the doc has not created the fallback behavior.
- P2 packages/realm/DESIGN.md:653: The sealed/evolution paragraph still lists AuthMethod and TokenEndpointAuthStrategy as sealed-add surfaces even though round-3 explicitly unsealed both. That stale text is not cosmetic: it contradicts the new chord principle and can mislead future maintainers into re-sealing extension points or treating open plugin families as changelog-managed closed sets.
- P2 packages/realm/DESIGN.md:848: Federation says v2 introduces FederatedRoomRef, but the code directly below already includes FederatedRoomRef at packages/realm/DESIGN.md:858 while the prose also says “v1 only constructs LocalRoomRef. v2 introduces FederatedRoomRef” at packages/realm/DESIGN.md:865. That is internally inconsistent. If the sealed family contains FederatedRoomRef in v1, v2 is not adding that subtype; if v2 adds it later, it cannot appear in the v1 engine contract. Pick one. The Carnot efficiency loss here is avoidable: the sealed surface is either reserved in full or grown later with a known break.

**The Good:**
- Round-2 PII closure mostly holds: packages/realm/DESIGN.md:419 narrows watchInRoom/watchFromFoyer to FullProjection/PublicProjection, and packages/realm/DESIGN.md:440 keeps PeerPresence empty, so the base type is not a covert PII channel.
- joinedAt removal from PublicProjection holds at packages/realm/DESIGN.md:465; the timing-surveillance rationale is explicit and structurally supported by the type shape.
- connectedTo removal from RoomDescriptor holds at packages/realm/DESIGN.md:242; the authorization surface is named and deferred to FederationGraphStore.
- AuthMethod and PreviewShape are now open extension points, which matches the architectural seal principle better than the earlier sealed designs.
- FoyerVisibility now has both parse and tryParse at packages/realm/DESIGN.md:318 and packages/realm/DESIGN.md:329, which closes Kelvin’s trust-boundary complaint.

**The Concerns:**
- LeaveReason and FoyerVisibility being enums is acceptable because they are engine-owned closed vocabularies, but the doc should stop calling enum additions additive in any casual sense. It does acknowledge analyzer breaks, which is the right posture.
- PeerPresence and RoomPreview being sealed still match architectural seals: audience projection and XOR preview shape are engine invariants, not plugin vocabularies.
- RasterPreview/VectorPreview/EmptyPreview are intentionally closed alternatives for preview transport, while PreviewShape is open inside VectorPreview. That split is coherent.
- RoomRef needs the v1/v2 contradiction fixed. The principle is sound, but the current text says both “reserved now” and “introduced later.”
- Round-1 additive-v2 overstatement is improved but not fully purged; there are still places where “minor-version bump” is used for analyzer-breaking changes. Hamming’s lesson is apt: the purpose of computing is insight, not numbers. SemVer labels should not hide source breaks.

nickmeinhold:
## Round-5 spiral closed → name the enforcing artifact

Pushed \`eb9fc904\`. Round-5 chord: **every claim of enforcement must name the artifact that does the enforcing.** Carnot's five round-4 findings all rhymed into "claim made, artifact contradicts" — aspirational discipline left as if it were enforced.

Generativity score: **3** (refinement of existing principles, but the four-chord series on PR #475 — see below — is the higher-leverage meta-artifact).

### The four-chord series on this PR

| Round | Chord | Catches |
|---|---|---|
| 1 | doc_violates_own_discipline | adjacent prose contradicts itself |
| 2 | discipline_in_types_not_prose | prose claims a discipline the return types don't enforce |
| 3 | seal_matches_architecture | types enforce a discipline the architecture disclaims |
| 5 | **name_the_enforcing_artifact** | architecture claims enforcement, but no real artifact exists at the layer below |

The series performs a single deeper discipline: **layered contract honesty.** Each layer's claim must point at the layer below that actually performs the discipline. Prose → types. Types → architecture. Architecture → build/package/runtime. The chain bottoms out at something the build system or runtime can verify; a claim that doesn't terminate at a real artifact is *aspirational, not enforced*.

### Cold termini

| # | Finding | Closure |
|---|---|---|
| F1 | SignedRequest moved to \`examples/\` but examples/ wasn't a real package | Doc now names the gap explicitly. Two v1-blocker artifacts added to migration plan: step 2 creates \`examples/livekit-token-server/\` as a workspace member with its own pubspec; step 4 wires \`dart pub deps --json\` whitelist in CI to fail the build on any signing/HMAC primitive in the engine package's transitive deps. Until those land: intentional aspirational discipline, cage-match grep is the compensating control. |
| F2 | AuthProviderId opaque-by-design but constants like 'google' invite spoofing | New display-only policy: AuthProviderId is a display-and-routing hint, NOT a trust primitive. Trust lives at the credential-exchange endpoint. Consumer rule: display surfaces MAY read providerIds; access-control decisions MUST NOT. \`realm_lints\` rule tracked under open questions. |
| F3 | PreviewShape "falls back to placeholder" with no defined semantics | Foyer fallback contract pinned: (1) per-shape skip, not per-preview discard; (2) telemetry event UnknownPreviewShapeEncountered, not error; (3) no exception crosses the foyer boundary. Binding on every \`PresenceService.watchFromFoyer\` consumer. |
| F4 | Versioning section still listed AuthMethod / TokenEndpointAuthStrategy as sealed-add surfaces | Rewrite distinguishes **audience-bounded sealed** (LeaveReason, FoyerVisibility, PeerPresence, RoomPreview, RoomRef — managed-breaking) from **plugin extension interfaces** (AuthMethod, TokenEndpointAuthStrategy, PreviewShape, branded IDs — additive-non-breaking). The two categories must never be confused. |
| F5 | RoomRef sealed family contained FederatedRoomRef while prose said v2 adds it | Resolved via **reserve-in-the-family** pattern: v1 declares both subtypes from day one; v1 only constructs LocalRoomRef; v1 consumers' exhaustive switches must already include a FederatedRoomRef arm. v2 lights up FederationGraphStore emission without changing the sealed surface — no breaking change. Same pattern applied to LeaveReason.portalTransit. The federation summary rewrites accordingly; only true v2 breaking-but-bounded change is RoomDescriptor.connectedTo field-add. |

### Engineering termini named

- **F1**: \`dart pub deps --json\` whitelist in CI (migration step 4) — structural enforcement that engine package's transitive deps reject signing/HMAC primitives.
- **F2**: \`realm_lints\` rule flagging any \`providerIds.contains(...)\` usage in trust-relevant code paths.
- **F3**: contract test feeding the reference foyer a VectorPreview with one known + one unknown shape, asserting per-shape skip + telemetry event.
- **F4 + F5**: doc-lint regex that cross-references prose mentions of sealed-as-X against actual code declarations and against sealed-family member lists — catches both stale text after spirals and reserve-vs-add-later contradictions.

### Operational notes

- **The series of five chords on one PR is anomalous.** Five rounds of cage-match producing five chord principles, each building on the prior, in a way that compounds. May warrant a global \`concept_layered_contract_honesty.md\` if the pattern reproduces on another PR — one more data point will tell us whether this is "what design-pin documents reliably produce under spiral discipline" or "particular to this PR's scope."
- **The doc is now 1034 lines.** A "navigating this document" TOC near the top is increasingly worth its cost.
- **F1's enforcement artifacts are v1 BLOCKERS, not v0 niceties.** The design pin commits the team to wire \`examples/livekit-token-server/\` as a workspace member and the deps whitelist into CI before v1 ships. If they're not in place at migration steps 2 and 4 as written, F2's package-boundary claim remains aspirational and the design pin is dishonest about its own enforcement story.

Audit at \`~/.claude/consolidation/2026-06-02T11-32-spiral/spiral-audit-PR475-round5.md\`.

Ready for round-5 cage-match (sixth cycle on this PR) whenever you call it. The chord principle should be increasingly verification-tractable as we descend — round-3 was a structural type change; round-5 is doc-prose precision. Round-6 cage-match should either find very small residuals (the pattern of each spiral landing smaller findings continues), OR find a meta-finding that the spiral series itself has reached a fixed point worth merging.

nickmeinhold:
## CarnotCodeCarver's Review

**Verdict:** REQUEST_CHANGES

**Summary:** Round 5 closed most of the original thermodynamic leaks, but the fixed-point test is not yet at equilibrium. The new chord I see is: enforcement_language_must_bind_to_the_actual_consumer. Two new contradictions remain in DESIGN.md.

**Findings:**
- packages/realm/DESIGN.md:535: The Foyer fallback contract is attached to the wrong interface. It says the rules are binding on every `PresenceService.watchFromFoyer` consumer, but `watchFromFoyer` returns `Stream<Set<PublicProjection>>`, whose only renderable room-presence payload is `opaqueAvatarRef` (`packages/realm/DESIGN.md:391`, `packages/realm/DESIGN.md:423`). `PreviewShape` only appears under `World.previewSnapshot()` via `VectorPreview` (`packages/realm/DESIGN.md:485`, `packages/realm/DESIGN.md:499`). A consumer could obey the entire `PresenceService` contract and never touch `PreviewShape`. Move the binding to `RoomPreview` / `World.previewSnapshot()` consumers, or explicitly define the foyer aggregation contract that consumes both `watchFromFoyer` and `previewSnapshot()`. As written, the enforcing artifact is named, but it is not the artifact carrying the value.
- packages/realm/DESIGN.md:171: The AuthProviderId section claims the `realm_lints` rule is “tracked under open questions,” but the Open questions list does not contain it (`packages/realm/DESIGN.md:801`). That reintroduces the round-5 failure mode: an enforcement claim points at a non-existent tracking artifact. Either add a concrete open question/blocker for `realm_lints` with scope, or remove the tracking claim and label it aspirational in the same disciplined way the SignedRequest section does.

**The Good:**
- F1 SignedRequest is materially improved. `packages/realm/DESIGN.md:332` names the two structural artifacts, `packages/realm/DESIGN.md:340` labels the current state as aspirational, and migration steps 2 and 4 make those artifacts real blockers (`packages/realm/DESIGN.md:773`, `packages/realm/DESIGN.md:779`). That is honest heat-accounting, not hand-waving.
- F2 AuthProviderId now has a section-local boundary. A downstream reader who reads only that section sees the critical rule: display MAY read `providerIds`; trust/access control MUST NOT (`packages/realm/DESIGN.md:167`). The section is mostly isolated enough, except for the missing lint tracking artifact above.
- F4 versioning is much clearer. The audience-bounded sealed versus plugin-extension distinction at `packages/realm/DESIGN.md:624` largely resolves the previous sealed/open confusion.
- F5 reserve-in-the-family is architecturally coherent: `FederatedRoomRef` and `LeaveReason.portalTransit` are declared in v1 and emission is deferred (`packages/realm/DESIGN.md:910`, `packages/realm/DESIGN.md:942`). That matches the intended non-breaking v2 emission path.

**The Concerns:**
- The reserve-in-the-family prose still leaves operational semantics a little under-specified. `assert or skip` for `FederatedRoomRef` (`packages/realm/DESIGN.md:913`) and `treat as disconnect` for `portalTransit` (`packages/realm/DESIGN.md:945`) are plausible, but production Dart `assert` disappears outside checked mode. I would prefer one explicit Dart switch example for each reserved variant so consumers know whether to log, drop, fail closed, or degrade.
- The F1 “aspirational discipline” label does not create a general slippery slope because it names blockers and interim grep control. The risk is cultural: every future aspirational claim must follow that exact pattern. Hamming’s warning applies: the purpose of computing is insight, not prose comfort. If there is no artifact, say so plainly.
- No untouched round 1-4 inconsistency jumped out beyond the two findings above. The remaining drift is operational: the doc now often knows what artifact should enforce the claim, but a few claims still point one noun away from the actual enforcement surface.

nickmeinhold:
## Round-6 inline closures pushed → \`8800a32c\`

Five round-6 findings addressed inline per "Address all 5 + push + merge" decision. Carnot's sub-chord (\`enforcement_language_must_bind_to_the_actual_consumer\`) shaped every fix.

| # | Finding | Fix |
|---|---|---|
| F3 | PreviewShape contract bound wrong consumer | Rebound to RoomPreview/VectorPreview consumers (the foyer renderer surface, NOT PresenceService.watchFromFoyer which sees PublicProjection). Two real artifacts named: \`realm_test::vectorPreviewConformance\` test + \`RoomPreviewRenderer\` mixin. |
| AuthProviderId | realm_lints claim pointed at missing entry | Open Question #9 added with explicit scope (lint rules a/b/c for AuthProviderId, SignedRequest, FederatedRoomRef). Doc comment now points at OQ#9(a) precisely. |
| Numbering leak | "F2 enforcement artifact" labels from round-4 | Scrubbed from migration steps 2 + 4. Rephrased as "first/second structural artifact." |
| PreviewShape "100% prose" | Kelvin's structural-enforcement upgrade | Two real artifacts named (conformance test + renderer mixin). Open Question #10 added. Honest claim-downgrade pattern: prose-grade until step 3 ships them; cage-match grep is the interim control. |
| FederatedRoomRef "MUST NOT" prose | Kelvin's structural-throw + my F5 Dart-pattern gap | Added \`assert\` in constructor gated on \`_federationActive\` flag that v2 flips. Dev/debug v1 fails loudly. Release covered by realm_lints rule (c). Canonical v1 consumer pattern pinned: fail-fast-with-telemetry (dispatch EngineInvariantViolation, then throw StateError) — same canonical pattern for LeaveReason.portalTransit. Silent absorption was what bit round-1; loud-by-default is the chosen default. |

The spiral series across 6 cycles + 5 chord principles formed a layered-contract-honesty discipline — each chord catches the previous chord's residue at the next layer of granularity. Round-6 found the next-finer sub-chord; round-6 inline closures applied it. The doc is shippable as the architectural pin.

Merging via \`gh pr merge --admin\` per session direction. Round-7 cage-match deferred to migration step 3 (engine interface PR) where the prose contracts begin meeting the analyzer.

PR #481, opened 2026-06-06
fix(audio): recompute proximity every frame + hysteresis + DF proximity signal

nickmeinhold:
## Why
Robin intermittently couldn't hear Nick until she moved/clicked. 

## Root cause
`BubbleManager`'s proximity loop — which drives per-peer audio enable/disable — was behind an early-return that only fired when the **local** player changed grid cell. Remote players (and Dreamfinder, which wanders autonomously) move too, so a peer could walk back into range and stay inaudible until the local player happened to move. Removed the guard → recompute every frame (cheap at meetup scale; bubble transitions stay guarded, audio toggles are no-ops when unchanged).

## Also in here
- **Audio hysteresis**: enable ≤4, disable >5 (was a hard cut at ≤2 while bubbles showed at ≤5 — a 3-square see-but-can't-hear dead zone + boundary flapping).
- **DF proximity signal**: client publishes `df-proximity {near}` on enter/exit of DF's range so the deployed bot can gate whose speech DF answers (near OR named). New `LiveKitTopic.dfProximity` + `LiveKitService.publishDfProximity`.
- Tests: stale-recompute + hysteresis regression test; LiveKitService fakes updated.

## Status
Held for review (per request) — green locally (`flutter analyze` + `flutter test` on touched files). Pairs with embodied-dreamfinder#58 (already deployed). A **distance fade** (Web Audio GainNode) is coming as a follow-up commit on this branch.

### DIFF
--- a/lib/flame/bubble_manager.dart
+++ b/lib/flame/bubble_manager.dart
@@ -104,6 +104,9 @@ class BubbleManager {
   final Map<String, PositionComponent> _playerBubbles = {};
   final Map<String, Vector2> _bubbleDisplacements = {};
   final Set<String> _audioEnabledParticipants = {};
+  /// Last volume pushed to LiveKit per participant, so the per-frame fade only
+  /// writes when the value actually changes (distance is an int → rarely).
+  final Map<String, double> _audioVolumes = {};
   /// Last reported local-player proximity to Dreamfinder, so the df-proximity
   /// signal is published only on enter/exit transitions, not every frame.
   bool _wasNearDreamfinder = false;
@@ -298,16 +301,12 @@ class BubbleManager {
     }
 
     // Notify Dreamfinder when the local player enters/exits its range so the
-    // bot can gate whose speech it hears. DF is in [nearbyPlayerIds] exactly
-    // when within visual range (the same threshold that shows its bubble), so
-    // "DF can hear you" lines up with "DF's bubble is up for you". Published on
-    // transition only — the bot holds the state between signals.
-    final nearDf = dreamfinderComponent != null &&
-        nearbyPlayerIds.contains(dreamfinderIdentity);
-    if (nearDf != _wasNearDreamfinder) {
-      _wasNearDreamfinder = nearDf;
-      _liveKitService?.publishDfProximity(near: nearDf);
-    }
+    // bot can gate whose speech it hears. null distance == DF not present.
+    _updateDreamfinderProximity(
+      dreamfinderComponent == null
+          ? null
+          : chebyshevDistance(playerGrid, dreamfinderComponent!.miniGridPosition),
+    );
 
     // Remove bubbles for players no longer nearby.
     final toRemove = <String>[];
@@ -476,6 +475,13 @@ class BubbleManager {
     _mergedBubble?.removeFromParent();
     _mergedBubble = null;
     _audioEnabledParticipants.clear();
+    _audioVolumes.clear();
+    // Emit a final exit so Dreamfinder doesn't hold a stale near:true after we
+    // tear down while the player was in range (cage match PR #481 — Carnot).
+    // Best-effort; the bot also self-heals on our ParticipantDisconnected.
+    if (_wasNearDreamfinder) {
+      _liveKitService?.publishDfProximity(near: false);
+    }
     _wasNearDreamfinder = false;
     _liveKitService = null;
     _dreamfinderAvatarBridge?.dispose();
@@ -645,6 +651,12 @@ class BubbleManager {
   }
 
   void _updateParticipantAudio(String playerId, int distance) {
+    // No LiveKit service yet (pre-connect / post-teardown) → don't mutate the
+    // gate state. Latching a state change we couldn't actually send leaves the
+    // gate stuck once the service comes up (the next frame sees "no change").
+    final service = _liveKitService;
+    if (service == null) return;
+
     final hasAudio = _audioEnabledParticipants.contains(playerId);
 
     // Hysteresis: enable when within the (tighter) enable threshold, disable
@@ -656,7 +668,7 @@ class BubbleManager {
 
     if (shouldEnable) {
       _audioEnabledParticipants.add(playerId);
-      _liveKitService?.setParticipantAudioEnabled(playerId, true);
+      service.setParticipantAudioEnabled(playerId, true);
       if (avDiagnosticsEnabled) {
         dispatch([AvAudioGateChanged(
           participant: playerId,
@@ -666,7 +678,8 @@ class BubbleManager {
       }
     } else if (shouldDisable) {
       _audioEnabledParticipants.remove(playerId);
-      _liveKitService?.setParticipantAudioEnabled(playerId, false);
+      _audioVolumes.remove(playerId); // re-set volume on next enable
+      service.setParticipantAudioEnabled(playerId, false);
       if (avDiagnosticsEnabled) {
         dispatch([AvAudioGateChanged(
           participant: playerId,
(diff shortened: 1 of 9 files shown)


nickmeinhold:
Distance fade landed (commit `f3836a80`).

**How the web fade works:** livekit_client 2.7.0 has no per-track volume API, so this sets `volume` on the `HTMLAudioElement` the SDK auto-creates per remote track (DOM id `livekit_audio_<cid>`). Conditional export keeps `package:web`/`dart:js_interop` out of native builds (typed interop, WASM-safe); native is a no-op for now (gate still applies).

**Curve:** full volume ≤1 square, linear ramp to silence at the disable threshold (5). The enable/disable subscription stays as the outer boundary + bandwidth saver.

**Review attention:** (1) the `livekit_audio_` id prefix is an internal livekit_client detail — degrades to no-op (full volume via gate) if it ever changes; (2) volume is set every frame while subscribed — cheap getElementById at meetup scale, could memoize if it ever shows up in a profile.

Tests: volume 0.5 @ d=3, full @ d=1; stale-recompute + hysteresis regression tests. Still held for your review.

nickmeinhold:
## CarnotCodeCarver's Review

**Verdict:** REQUEST_CHANGES

**Summary:** One correctness leak in the new DF proximity signal keeps this from Carnot efficiency. As Dijkstra would put it, correctness is a precondition, not an afterthought; a reliable enter/exit protocol must cover teardown and identity-loss paths, not just steady-state frames. I did not run tests, per instruction; I verified by reading repo code, pubspec.lock, and livekit_client 2.7.0 package source.

**Findings:**
- lib/flame/bubble_manager.dart:307 publishes `df-proximity` only from the per-frame transition check, but lib/flame/bubble_manager.dart:452 and lib/flame/bubble_manager.dart:479 can drop/reset Dreamfinder proximity state without sending `near:false`. The PR states the bot holds state between signals; if the local player was near DF and the manager clears, Dreamfinder leaves, or the DF identity/component is reset before another update observes `nearDf == false`, the bot can retain a stale `near:true` gate. Add a small transition helper that publishes exit before clearing/changing DF state while `_wasNearDreamfinder` is true, and cover it with a regression test.

**The Good:**
- lib/flame/bubble_manager.dart:202 correctly removes the local-player-grid early return; the old optimization was thermodynamically lossy because remote-player/DF motion changed the system while the cached predicate observed only the local coordinate.
- lib/flame/bubble_manager.dart:654 uses explicit hysteresis state instead of a single threshold, which is the right control-system shape for boundary flapping.
- lib/livekit/livekit_topic.dart:12 keeps topics as a Dart enum with wire values, a better Dart 3 shape than scattering raw strings.
- lib/livekit/set_track_volume.dart:13 uses a conditional export so `package:web`/`dart:js_interop` stay out of native builds.
- lib/livekit/set_track_volume_web.dart:23 matches livekit_client 2.7.0 internals: pubspec.lock resolves livekit_client 2.7.0, and /Users/nick/.pub-cache/hosted/pub.dev/livekit_client-2.7.0/lib/src/track/web/_audio_html.dart:25 declares `audioPrefix = 'livekit_audio_'`; /Users/nick/.pub-cache/hosted/pub.dev/livekit_client-2.7.0/lib/src/track/remote/audio.dart:47 passes `getCid()` to `startAudio`.

**The Concerns:**
- pubspec.yaml:12 still allows `livekit_client: ^2.5.1` while lib/livekit/set_track_volume_web.dart:5 depends on an internal DOM id scheme verified only for the locked 2.7.0 source. That is acceptable for a committed lockfile app, but it deserves a loud upgrade checklist or a narrow SDK constraint because the failure mode is silent loss of fade.
- lib/flame/bubble_manager.dart:684 does a DOM lookup through LiveKitService every frame for every subscribed participant on web. Meetup-scale is probably fine, but this is a heat leak: cache last sent volume per participant or quantize updates if participant counts grow.
- test/flame/bubble_manager_test.dart:293 covers the happy-path fade values, but there is no test for `df-proximity` enter/exit, teardown exit, or Dreamfinder removal. The new protocol is stateful and reliable; it needs lifecycle tests, not only fakes updated at test/chat/chat_service_test.dart:1165 and test/map_editor/map_sync_service_test.dart:623.

nickmeinhold:
## Cage-match findings addressed (commit `e65e8414`)

Both adversarial reviewers blocked on the same shape — the `df-proximity` signal was a fragile per-frame transition latch. Fixed as one chord:

| Finding | Reviewer | Fix |
|---|---|---|
| Null-`_liveKitService` latches the boolean but never sends → signal lost forever | Kelvin | Don't latch `_wasNearDreamfinder` when service is null; re-fires next frame. Same guard added to the human audio gate. |
| Teardown/reset can drop DF state without an exit → bot holds stale `near:true` | Carnot | `clear()` now publishes `near:false` if the player was near DF. |
| DF signal had no hysteresis (human audio did) → boundary spam on reliable channel | Maxwell | DF enter ≤4 / exit >5, matching the audio gate. |
| Volume re-written every frame even when unchanged | Kelvin + Maxwell | `_audioVolumes` memoizes; fade only writes on change. |
| "smooth gradient" comment vs integer steps | Kelvin | Reworded to "per-square ramp / stepwise". |

Extracted `_updateDreamfinderProximity(int?)` with a `@visibleForTesting` seam + **4 regression tests** (hysteresis, DF-absent exit, null-service no-latch, clear() exit). All 41 bubble_manager tests green; `flutter analyze` clean.

Re-request review when ready.

nickmeinhold:
## CarnotCodeCarver's Review (re-review)

**Verdict:** REQUEST_CHANGES

**Summary:** The specific re-review target is mostly fixed: `clear()` now emits the final `df-proximity near:false`, and the null-service latch no longer traps a lost enter. However, the refactor introduced a new volume-cache race in the added fade path, so I would not approve yet. As Dijkstra would put it, testing shows presence, not absence; here the state proof fails before tests run.

**Findings:**
- P2 `lib/flame/bubble_manager.dart:699` + `lib/livekit/livekit_service.dart:502`: `_audioVolumes[playerId]` is cached before `LiveKitService.setParticipantAudioVolume()` proves it actually touched a track. If the participant is in range while their audio publication exists but `publication.track` is still null, `setParticipantAudioVolume()` no-ops at line 504, but BubbleManager records the volume at lines 700-702. When the audio track later subscribes at `lib/livekit/livekit_service.dart:846`, there is no hook to replay the cached proximity volume. Result: the new track can play at default volume 1.0 indefinitely while distance is unchanged. This is a new regression from the fade refactor. Either return success/failure from `setParticipantAudioVolume`, avoid caching when no track was touched, or replay volume on `TrackSubscribedEvent`.
- P3 `test/flame/bubble_manager_test.dart:293`: the new fade test only covers the steady-state case where the mocked service accepts volume immediately. It does not cover the late-track/no-op path above, so the test suite would still go green with the bug present. Add a regression around publication track absence or a LiveKit track-subscribed replay seam.

**The Good:**
- `lib/flame/bubble_manager.dart:482`: the stale DF teardown issue from the prior review is fixed in the intended order: publish `near:false` while `_liveKitService` is still reachable, then reset `_wasNearDreamfinder` and clear the service.
- `lib/flame/bubble_manager.dart:731`: Kelvin’s null-service latch is handled correctly. The transition is compared, then the service null check returns before mutating `_wasNearDreamfinder`, so the same near distance can fire once service is set.
- `lib/flame/bubble_manager.dart:666`: remote proximity is now recomputed every frame and the audio hysteresis state machine is explicit. This addresses the original remote-moves-while-local-still failure mode without stringly state.
- `lib/livekit/livekit_topic.dart:50`: adding `dfProximity` to the existing `LiveKitTopic` enum is the right Dart 3 direction; the wire string is localized instead of scattered.

**The Concerns:**
- `clear()` still fire-and-forgets `publishDfProximity` at `lib/flame/bubble_manager.dart:483`. Given the stated best-effort teardown semantics this is acceptable, but it is not a hard delivery guarantee if a caller immediately tears down the room.
- The new web fade reaches into LiveKit’s internal DOM id scheme at `lib/livekit/set_track_volume_web.dart:23`. The failure mode is a no-op, not a crash, so this is tolerable, but it should be treated as an SDK-coupled affordance rather than a stable API. Feynman’s warning applies: if the model of the machinery is wrong, nature will still do what it does.

nickmeinhold:
## CarnotCodeCarver's Review (round 3)

**Verdict:** REQUEST_CHANGES

**Summary:** The BubbleManager-level retry fix is correct for `publication.track == null`: `_audioVolumes` is only cached after `setParticipantAudioVolume` returns true. However, the bool contract is still too high in the stack. It returns true when a LiveKit track object exists, not when the web audio element was actually found and written. As Dijkstra put it, testing can show the presence of bugs, not their absence; here the regression test mocks the service and does not exercise the colder sink where the entropy leaks.

**Findings:**
- High: `setParticipantAudioVolume` can still cache after a no-op DOM write. At `lib/livekit/livekit_service.dart:512`, the service calls `setTrackVolume(track.getCid(), volume)` and then sets `applied = true` solely because `publication.track` is non-null. But the web implementation at `lib/livekit/set_track_volume_web.dart:24`-`25` can return without setting volume when the `HTMLAudioElement` is not present or is not an audio element. Because `setTrackVolume` is `void`, that miss is reported as success to `BubbleManager` at `lib/flame/bubble_manager.dart:709`-`710`, which caches the volume and suppresses retries. This is the same thermodynamic irreversibility in a different chamber: state says work was done, but no heat reached the reservoir. Make `setTrackVolume` return `bool` on web and have `setParticipantAudioVolume` return true only if at least one volume write actually landed, or otherwise avoid caching web misses.

**The Good:**
- `lib/flame/bubble_manager.dart:705`-`710` correctly defers `_audioVolumes` caching until the service reports success, fixing the null-track flavor of the late-subscription bug.
- `lib/flame/bubble_manager.dart:458`-`463` cleans up both `_audioEnabledParticipants` and `_audioVolumes` in `removeBubble`, which closes the stale map-entry leak for vanished peers.
- `test/flame/bubble_manager_test.dart:320`-`348` covers the intended retry behavior at the BubbleManager/service boundary.

**The Concerns:**
- I did not run tests, per instruction.
- The new regression test mocks `setParticipantAudioVolume`; it verifies BubbleManager’s caching policy, but not LiveKitService’s bool accuracy or the web DOM miss path.

nickmeinhold:
## CarnotCodeCarver's Review (round 4)

**Verdict:** APPROVE

**Summary:** Round-4 confirmation: the previous irreversibility is closed. As Dijkstra put it, “Program testing can be used to show the presence of bugs, but never to show their absence”; with that caveat, this review finds no remaining blocker in the revised volume-application path. I did not run tests, per instruction.

**Findings:**
- None — prior irreversibility closed.

**The Good:**
- [lib/flame/bubble_manager.dart:705] now compares the desired volume against the cache, but [lib/flame/bubble_manager.dart:709] only updates that cache after `setParticipantAudioVolume` returns true. That preserves retry pressure until the effect is actually applied.
- [lib/livekit/livekit_service.dart:509] initializes `applied = false`, and [lib/livekit/livekit_service.dart:512] only flips it when `publication.track != null` and `setTrackVolume(track.getCid(), volume)` succeeds. This removes the prior false-success path where a LiveKit track existed but the web audio element was not ready.
- [lib/livekit/set_track_volume_web.dart:27] returns true only after finding an `HTMLAudioElement` and writing its clamped volume at [lib/livekit/set_track_volume_web.dart:30]. A missing or not-yet-appended element returns false at [lib/livekit/set_track_volume_web.dart:29], so the late DOM append is no longer stranded at default volume.
- [lib/livekit/set_track_volume_stub.dart:11] returning true is acceptable: on native platforms the fade is intentionally unsupported, so retrying forever would add heat without doing work. The Carnot accounting is correct here: no recoverable lower-temperature reservoir exists on native yet.
- [test/flame/bubble_manager_test.dart:320] still mocks `LiveKitService`, which is appropriate for the BubbleManager regression: the test verifies the cache/retry contract by forcing false, false, true from `setParticipantAudioVolume` and observing retries at [test/flame/bubble_manager_test.dart:337] through [test/flame/bubble_manager_test.dart:348].

**The Concerns:**
- Residual coverage gap, not a blocker: there is still no direct unit/browser test for `setTrackVolume` against an actual DOM element id at [lib/livekit/set_track_volume_web.dart:28]. The mocked BubbleManager regression proves the thermodynamic valve is reversible again at the service boundary, but not the SDK-internal id assumption itself.

PR #486, opened 2026-06-21
feat(audio): disable DF audio on subscribe + Room seam for silence tests (#20)

### LINKED ISSUE #20: Game is flashing a white screen at times
I determined the flashing only occurs when audio is detected and a user’s window is chosen (and the blue border appears)

nickmeinhold:
Addresses both MEDIUM follow-ups deferred from the PR #485 cage match (Carnot), tracked as task #20.

## (1) Deferred-component window — disable DF audio on subscribe unconditionally

DF audio is proximity-gated: `BubbleManager._updateDreamfinderAudio` is the **sole enabler** and only forwards DF audio when the local player is in range. But that gate doesn't run until DF's `DreamfinderComponent` exists, which is deferred to `gameReady` (it needs the path component). Previously the `TrackSubscribedEvent` handler only disabled a fresh DF audio track when **already silenced** — so between a DF track subscribing and the gate's first tick, an *unsilenced* fresh track was **audible from anywhere on the map**.

**Fix:** disable DF audio on subscribe **unconditionally** (`applyDreamfinderSilenceOnSubscribe` → `disableDreamfinderAudioOnSubscribe`, dropped the `dreamfinderSilenced.value` condition). The gate re-enables on the first near-frame.

**Why this is safe (latch analysis):** the gate tracks state in `_audioEnabledParticipants`, which **starts empty** — it already assumes the track begins disabled. Disabling on subscribe just makes the SFU match that assumption, so the gate's `shouldEnable = !hasAudio && near` transition still fires correctly when the player walks into range. No latch conflict.

**agent-* timing:** `dreamfinderIdentity` is reconciled to the real `agent-*` identity at participant-join (`tech_world.dart:527/537`), and the audio track belongs to that same participant — so the gate keys off the matching identity. Disabling an `agent-*` track on subscribe just means it's muted until the gate's first near-tick, which is the intended proximity-gated behaviour.

## (2) Room seam for `setDreamfinderSilenced`

It read `_room` directly, so its participant iteration wasn't unit-testable. Added a `@visibleForTesting remoteParticipantIdentities` provider (defaults to reading `_room.remoteParticipants`) alongside the existing `silenceParticipantAudio` seam. `setDreamfinderSilenced` now iterates via the seam.

**Production behaviour unchanged:** the default provider returns `const []` when `_room` is null, so the old null-guard early-return is preserved as a harmless no-op.

## Tests — `test/livekit/livekit_service_df_silence_test.dart` (10 cases)

**Saw-it-fail-then-pass confirmed** for both behaviour changes (reverted impl to the old conditional → the unconditional-disable cases failed; restored → green).

Disable-on-subscribe:
- DF audio (not silenced) → disables ✅ · DF audio (silenced) → disables ✅ · `bot-dreamfinder` → disables ✅
- DF **video** → no-op ✅ · non-DF bot (`bot-claude`) → no-op ✅ · human peer → no-op ✅

`setDreamfinderSilenced`:
- `(true)` disables every DF identity in the roster (`bot-dreamfinder` + `agent-*`), nobody else ✅
- `(true)` with no DF in room → nobody disabled ✅
- `(false)` touches nobody — the #485 un-silence-no-force-enable leak, now pinned at the **service** level (was only pinned at the gate level + read-verified on the button) ✅
- `(true)`→`(false)` → no further disable, no enable ✅

Also adds a catch-all `noSuchMethod` to two hand-rolled `implements LiveKitService` fakes (`chat_service_test`, `map_sync_service_test`) that predate the `extends Mock` convention — adding any public interface member broke them.

## Verification
- `flutter analyze --fatal-infos` — clean over `lib/` + `test/` (only reported issues are inside the gitignored `build/macos/SourcePackages/` firebase bundled tests; CI never sees them).
- `flutter test` — all **1945** pass.

## Notes
- **Last gate (live):** the deferred-component-window fix is repro-blocked on a live DF session (harness #3) — Nick walking up to a freshly-joined DF and confirming audio fades in correctly. Logic + unit coverage are in; the experiential gate is not yet cleared.
- **Review tier: cage-match** — touches LiveKit track lifecycle + the DF-silence trust/state surface with prior leak history (#485/#594), and (1) is a genuine behaviour change.

Closes #20.

### DIFF
--- a/lib/livekit/livekit_service.dart
+++ b/lib/livekit/livekit_service.dart
@@ -102,17 +102,17 @@ class LiveKitService {
     @visibleForTesting Future<String?> Function()? tokenRetriever,
     @visibleForTesting void Function(String identity)? silenceParticipantAudio,
     @visibleForTesting Iterable<String> Function()? remoteParticipantIdentities,
-  })  : _tokenRetriever = tokenRetriever,
-        // Seams for unit-testing the DF-silence logic without faking the
-        // LiveKit SDK. Default to the real server-side disable / the live
-        // `Room`'s participant set. The late-init dance is because the default
-        // closures capture `this`.
-        _silenceParticipantAudio = silenceParticipantAudio,
-        _remoteParticipantIdentities = remoteParticipantIdentities {
-    _silenceParticipantAudio ??=
-        (identity) => setParticipantAudioEnabled(identity, false);
-    _remoteParticipantIdentities ??=
-        () => _room?.remoteParticipants.values.map((p) => p.identity) ?? const [];
+  }) : _tokenRetriever = tokenRetriever {
+    // Seams for unit-testing the DF-silence logic without faking the LiveKit
+    // SDK. Default to the real server-side disable / the live `Room`'s
+    // participant set. Assigned in the constructor body (not the initializer
+    // list) because the default closures capture `this` — which lets the fields
+    // be `late final` rather than nullable, encoding "set once, never null".
+    _silenceParticipantAudio = silenceParticipantAudio ??
+        ((identity) => setParticipantAudioEnabled(identity, false));
+    _remoteParticipantIdentities = remoteParticipantIdentities ??
+        (() =>
+            _room?.remoteParticipants.values.map((p) => p.identity) ?? const []);
   }
 
   final String userId;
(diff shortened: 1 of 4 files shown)


nickmeinhold:
## 🥊 Cage-match: 3-way consensus APPROVE

- **Maxwell** (Claude): APPROVE
- **Kelvin** (Gemini 3 Pro): APPROVE
- **Carnot** (Codex GPT-5.5): APPROVE

Convergent non-blocking findings addressed inline (commit e8d3881a):
- seam fields nullable+bang → `late final` (3/3 consensus)
- `await service.dispose()` in test tearDowns (Carnot)

Deferred (acceptable, noted): verbose doc comment (Kelvin), `noSuchMethod` catch-all on hand-rolled fakes lowers future test sensitivity — real fix is migrate to `extends Mock` next time through.

Issue #17, opened 2024-07-17
Second client doesn't add first Player to the World


nickmeinhold:
I debugged and the NetworkingService is adding the new NetworkUser to the `userAdded` stream.

I suspect the _userAddedStream in TechWorld is being listened to _after_ the userAdded event has already been emitted.

- [ ] Try using a BehaviourSubject or a RelaySubject

PR #487, opened 2026-06-21
feat(movement): WASD v2 — continuous-while-held + diagonals (#17)

nickmeinhold:
## What

Upgrades the discrete keyboard movement from PR #483 to **continuous-while-held** with **diagonals**.

- **Hold to walk**: holding a movement key keeps the player stepping in that direction until released. `TechWorldGame` now tracks the pressed-key set in `onKeyEvent` and auto-repeats in `update(dt)`, instead of moving one cell per physical key-down.
- **Diagonals**: holding two perpendicular keys (e.g. `W`+`D`) moves diagonally; opposing keys on the same axis (`W`+`S`) cancel.
- **No √2 speed boost**: cadence is one cell-step per `PlayerComponent.cellMoveDuration` regardless of cardinal vs diagonal. `movementVelocity` additionally normalises magnitude so diagonal == cardinal — the pure, tested definition of the constraint.

## How it stays safe

Every step still routes through the existing `TechWorld.moveInDirection` → `movePlayerToCell` path (pathfind → collide → animate → broadcast over LiveKit). Keyboard movement inherits collision and remote position sync for free; there is no new wire path and no bypass of pathfinding.

The new logic is split into pure, Flame-free helpers in `keyboard_movement.dart` so each behaviour is unit-tested in isolation:
- `directionForKeys(Set<LogicalKeyboardKey>)` — combines held keys into one `Direction` (diagonals + axis cancellation).
- `movementVelocity(Direction, {speed})` — normalised velocity (no √2 boost).
- `MovementTicker` — cooldown-gated cadence; `tick(dt)` is driven exactly as the game loop would, giving an immediate first step then one per interval.

Move duration is lifted to `PlayerComponent.cellMoveDuration` so the auto-repeat interval and the per-cell animation share one source of truth.

## Tests

- **+18 new tests** (`test/flame/keyboard_movement_v2_test.dart`): combined-direction resolution incl. diagonals & cancellation, diagonal normalisation (magnitude == cardinal), and held-key cadence via `MovementTicker.tick(dt)`.
- ATDD: verified each behaviour's test **fails without its impl** (broke normalisation, ticker, and diagonal resolution in turn; confirmed the matching tests went red), then restored.
- Full suite: **1952 passing** (was 1934). `flutter analyze --fatal-infos` clean on tracked files.

## Lane

Touches only `lib/flame/shared/keyboard_movement.dart`, `lib/flame/tech_world_game.dart`, `lib/flame/components/player_component.dart`, and the new test. No `lib/livekit/**`, bubble/bridge, chat, or timer files.

### DIFF
--- a/lib/flame/components/player_component.dart
+++ b/lib/flame/components/player_component.dart
@@ -65,10 +65,10 @@ class PlayerComponent extends SpriteAnimationGroupComponent<Direction>
 
   /// Duration of a single one-cell [MoveToEffect].
   ///
-  /// Lifted to a named constant so the keyboard auto-repeat cadence
-  /// ([TechWorldGame]'s continuous-while-held tick) can stay in lock-step with
-  /// the per-cell animation: one held-key step is issued per [cellMoveDuration],
-  /// so a key fires exactly when the previous cell-move completes.
+  /// This animation duration *is* the continuous-keyboard-movement cadence:
+  /// [TechWorldGame] issues the next held-key step only once the previous cell
+  /// move has finished (gated on [isMoving]), so a held key walks at exactly one
+  /// cell per [cellMoveDuration] with no separate repeat timer.
   static const double cellMoveDuration = 0.2;
 
   List<MoveEffect> _moveEffects = [];
--- a/lib/flame/components/player_component.dart
+++ b/lib/flame/components/player_component.dart
@@ -160,6 +160,21 @@ class PlayerComponent extends SpriteAnimationGroupComponent<Direction>
         position.y.round() ~/ gridSquareSize,
       );
 
+  /// Whether a cell-move is currently animating.
+  ///
+  /// A [MoveToEffect] is added per path segment by [move] and auto-removed on
+  /// completion (`removeOnFinish` defaults to `true`), so the presence of any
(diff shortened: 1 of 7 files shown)


nickmeinhold:
## 🥊 Cage-match — findings resolved (merging)

Round 1: Maxwell APPROVE · **Kelvin errored (Gemini network ETIMEDOUT)** · **Carnot REQUEST_CHANGES** — 3 real P1s the author + Maxwell missed:
- **P1 cadence:** the isMoving-gate + MovementTicker combo produced stop-wait-step (~0.4s/cell), not continuous. Fixed by deleting the ticker — isMoving alone paces the walk (animation duration == cadence).
- **P1 focus-stranding:** focus checked only in onKeyEvent; mouse/programmatic focus while a key was held kept walking the avatar. Fixed with an isTextFieldFocused() guard at the top of update().
- **P1 diagonal/dead-code:** movementVelocity (the 'no √2 boost' proof) was test-only — runtime diagonals are √2 faster on screen. Deleted the dead helper, reworded to the honest grid-cadence truth; pixel-speed normalisation filed as a design call.

All 3 fixed (commit 190f9dd3), both proofs fail-then-pass verified, 1951 green, analyze clean.

Issue #23, opened 2025-12-27
Create an in-game IDE that you can solve coding challenges in


nickmeinhold:
Robin suggested: 
- an IDE with no code complete or ID
- but you can walk over the claude and ask questions, then walk back
- After each exercise you can choose "that was too easy" or "that was too hard"
- And if you get stuck you can click "This is too hard"
- its like [TMC](https://tmc.mooc.fi/) (test my code) - it has unit tests which test your code and if its correct you get a gold coin (I reckon points and levels)

---

## Sub-tasks
- [ ] #25 - Add code editor component to game world
- [ ] #26 - Create coding challenge system with unit tests  
- [ ] #27 - Add difficulty feedback system (too easy/too hard)
- [ ] #28 - Add 'I'm stuck' button with Claude bot integration
- [ ] #29 - Add points and rewards system for completed challenges

Issue #13, opened 2024-07-17
Listening to PlayersService's userAdded stream throws when new player arrives


nickmeinhold:
When listening to the stream we add a new PlayerComponent to the World but the World hasn't finished loading.

PR #488, opened 2026-06-21
feat(chat): DM quote-reply with display-only quote snapshot

nickmeinhold:
Adds quote-reply to DM threads (task #13, chat lane only). A reply
references the message it replies to via a nullable replyToMessageId
parsed defensively at the wire seam, plus a denormalized
replyToText / replyToSenderName snapshot so the quote renders even when
the original isn't in the local list.

Trust boundary preserved: the reply's senderId is still derived from the
LiveKit transport identity, never the wire payload. The quote snapshot is
display-only (cosmetic, like senderName) and never a trust anchor — a new
test asserts a reply with a spoofed payload senderId is still filed under
the transport identity. Malformed reply fields (wrong type) are coerced to
null rather than thrown, so a bad payload can't tear down the stream.

In-app notifications + unread badge already exist (totalUnreadNotifier,
ConversationListTile unread pill, markConversationRead); this round hardens
them with regression coverage rather than rebuilding.

UI: long-press or "Reply" hint on a DM bubble starts a reply; a "Replying
to X" banner shows above the input with a cancel button; reply bubbles
render the quoted snippet.

Out of scope (separate task #23): the on-map mailbox world-entity unread
indicator (Flame). No lib/flame/** touched.

Tests: model reply round-trip + defensive parse; service send/inbound
reply + forged-senderId-in-reply; new DmThreadView widget test for the
compose/cancel/publish flow. All green (flutter analyze lib test clean;
flutter test 1950 passing).

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/lib/chat/chat_message.dart
+++ b/lib/chat/chat_message.dart
@@ -8,10 +8,18 @@ class ChatMessage {
     this.participants,
     this.isLocalUser = false,
     this.isBot = false,
+    this.replyToMessageId,
+    this.replyToText,
+    this.replyToSenderName,
     DateTime? timestamp,
   }) : timestamp = timestamp ?? DateTime.now();
 
   /// Reconstruct a [ChatMessage] from a Firestore document map.
+  ///
+  /// Optional fields are parsed defensively with Dart 3 if-case patterns
+  /// rather than blind `as` casts: a malformed value (wrong type, e.g. an int
+  /// where a string is expected) drops the field instead of throwing. This
+  /// matters at the wire/persistence seam where the payload is untrusted.
   factory ChatMessage.fromFirestore(Map<String, dynamic> json) {
     return ChatMessage(
       text: json['text'] as String? ?? '',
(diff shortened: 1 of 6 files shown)


nickmeinhold:
## 🥊 Cage-match round 2 — findings resolved

**Round 1:** Maxwell APPROVE · Kelvin APPROVE · **Carnot REQUEST_CHANGES** (caught a HIGH the other two missed).

**Carnot's HIGH:** `_fetchAndCacheHistory` dropped the reply fields on history rehydration → persisted replies reverted to plain messages after reload. The model round-trip test passed; the bug lived at the *service* re-construct copy-site one layer up. (The group copy-site had the same latent bug — both fixed.)

**All 4 findings fixed (commit d905f189), verified:**
- HIGH: reply fields carried through BOTH DM + group copy-sites; regression tests assert on the **service path** (loadHistory), saw-it-fail confirmed.
- MEDIUM: `fromFirestore` made total — `_parseParticipants` + `_parseTimestamp` fallback, 6 malformed-doc tests (incl. a fully-malformed legacy doc).
- LOW: half-reply API eliminated — `sendDm` takes only `replyTo`, derives id+snapshot together (half-reply unrepresentable).
- NIT: duplicate `_asStringOrNull`/`_stringOrNull` → one `ChatMessage.asStringOrNull`.

1958 tests green · analyze clean. Carnot's blocking HIGH resolved → clearing the hold.

PR #489, opened 2026-06-21
feat(timer): late-joiner catch-up + in-world clock entity (#19)

nickmeinhold:
## Timer v2

Two upgrades to the shared room countdown shipped in #484.

### 1. Late-joiner catch-up (primary — pure correctness)
A participant joining a room **after** a countdown started previously began a fresh full-duration countdown. Now they catch up to the authoritative remaining time.

- `CountdownTimerState` models the countdown against an **absolute end instant**, not a decrementing counter. `remaining` is always re-derived as `endsAt - now()` through an **injectable clock** (DI seam — no wall-clock in tests). New `startAt(DateTime endsAt)` is the late-joiner entry point.
- `TimerService` reconstructs `endsAt = startedAtMillis + duration` from the broadcast `start` message, so a joiner mid-countdown sees the real remaining (e.g. ~40s of a 60s timer started 20s ago), not 60s.
- **No wire change** — the v1 author already shipped `startedAtMillis` on the wire as a hook for exactly this. Messages without a timestamp fall back to a fresh local countdown. An already-expired timer finishes immediately on receipt (alarm + banner), matching an always-present client.
- Bonus: the absolute-end model is self-correcting against skipped ticks (backgrounded tab) — the next tick snaps to the true remaining rather than drifting.

### 2. World-entity clock rendering
- New `CountdownClockComponent` (`lib/flame/components/`) renders the countdown as a positioned Flame entity **in the world** — a wall clock the whole room can read — not only the UI overlay. It's a pure **view** of `CountdownTimerState`: owns no logic, never mutates. Hidden when idle; `mm:ss` while running; a `TIME'S UP` face while the alarm is active. y-priority depth sorting like doors/players.
- `TechWorld` adds the clock on LiveKit connect at a default position near the map spawn, removes it on disconnect.

### Design gate
Map-editor **placement** of the clock (placing/moving via the editor) is **deferred / design-gated, owned by Nick** — not built here. The clock renders at a sensible default near spawn.

## Tests
- 57 timer + clock tests; **full suite 1950 green**; `flutter analyze --fatal-infos` clean on `lib`/`test`.
- Late-joiner catch-up covered explicitly with a deterministic injected clock (start at T0 with end=T0+120, join at T0+30 → asserts 90s). **Saw-it-fail-then-pass confirmed**: the catch-up test fails when the reconstruction logic is reverted, passes with it.

## Review tier
Recommend **cage-match**: this touches wire-format interpretation + cross-participant state-sync (clock skew, expired-on-arrival, missing-timestamp fallback) — exactly the multi-participant invariants a different-inductive-bias adversary catches.

### DIFF
--- a/lib/flame/components/countdown_clock_component.dart
+++ b/lib/flame/components/countdown_clock_component.dart
@@ -98,18 +98,25 @@ class CountdownClockComponent extends PositionComponent {
     super.onRemove();
   }
 
-  @override
-  void render(Canvas canvas) {
-    if (!_visible) return;
+  // Cached laid-out paragraph + the (label, finished) key it was built for.
+  // The clock renders every frame but its text only changes ~once a second, so
+  // we rebuild the ui.Paragraph only when the label or alarm face changes.
+  ui.Paragraph? _cachedParagraph;
+  String? _cachedLabel;
+  bool? _cachedFinished;
 
-    final finished = !state.running && alarmActive.value;
-    final rect = Rect.fromLTWH(2, 2, size.x - 4, size.y - 4);
-    final rrect = RRect.fromRectAndRadius(rect, const Radius.circular(4));
+  /// Number of times a ui.Paragraph was actually (re)built — exposed so a test
+  /// can prove the cache rebuilds only on label/face change, not every frame.
+  @visibleForTesting
+  int paragraphBuildCount = 0;
 
-    canvas.drawRRect(rrect, finished ? _alarmBgPaint : _bgPaint);
-    canvas.drawRRect(rrect, finished ? _alarmBorderPaint : _borderPaint);
-
-    final label = finished ? "TIME'S UP" : state.formatted;
+  ui.Paragraph _paragraphFor(String label, bool finished) {
+    if (_cachedParagraph != null &&
+        _cachedLabel == label &&
+        _cachedFinished == finished) {
+      return _cachedParagraph!;
+    }
+    paragraphBuildCount++;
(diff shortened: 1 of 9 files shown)


nickmeinhold:
## 🥊 Cage-match — BLOCKER + findings resolved

Round 1: Kelvin APPROVE · Maxwell + **Carnot REQUEST_CHANGES**. Carnot caught that late-joiner catch-up — the feature's whole point — **wasn't actually delivered**: TimerService only listened to future messages and nothing republished the timer on participant-join (LiveKit doesn't replay), so a mid-countdown joiner saw idle state. The tests injected the start message directly, proving the timestamp math but not the delivery.

**All findings fixed (commit 35542657), verified:**
- **BLOCKER:** `republishForJoiner()` wired into the participant-joined handler (mirrors `publishMapInfo`). Tested on the real stream→_apply path (not direct injection).
- **HIGH (clock skew):** wire now carries *relative* remaining-seconds, anchored to the receiver's own clock (`endsAt = now + remaining`) — skew-immune by construction. Tested with divergent sender/receiver clocks.
- **MEDIUM (start/cancel race):** monotonic `generation` on every message; `_apply` drops stale generations and refuses to resurrect a cancelled generation. Tested (stale-cancel-after-newer-start ignored).
- **LOW:** clock component caches its laid-out paragraph, rebuilds only on label change.
- **NIT:** dead `step` param removed from `tick`.

68 timer/clock tests green · analyze clean. Verified the new republish + generation-guard logic by reading. Clearing the hold.

PR #498, opened 2026-06-24
fix(web): auto-recover from WebGL context loss instead of freezing

nickmeinhold:
## What happened

Andy's video froze mid-meetup while audio kept working (Firefox 152, M3 Max). The console showed the real cause:

```
WARNING: Falling back to CPU-only rendering. Reason: failed to create GrContext. Error: RuntimeError: index out of bounds
Uncaught RuntimeError: index out of bounds   (canvaskit.wasm)
Another exception was thrown: Instance of 'minified:kG<void>'   ×76 → ×921 ...
```

This is **not a video-track problem** — it's a **WebGL context loss**. Flutter's CanvasKit renderer draws the *entire* game (map, avatars, video bubbles) through one WebGL context. When the browser drops that context, CanvasKit can't rebuild its `GrContext`, falls back to CPU rendering, which *also* throws — and the whole canvas freezes on its last good frame. **Audio survives because LiveKit audio is a completely separate pipeline that never touches CanvasKit.** "Video froze, audio fine" is the fingerprint.

The suspected trigger is GPU memory pressure from the per-frame video-bubble texture churn (`decodeImageFromPixels` → new `ui.Image` every frame, per bubble), which Firefox is more prone to dropping a context under.

## The gap

There was **no context-loss recovery anywhere in the app**. A transient GPU hiccup became a permanent silent freeze, recoverable only if the user happened to know to hard-reload.

## The fix

Flutter 3.44 exposes no in-place CanvasKit re-init, so this adds a capture-phase `webglcontextlost` watchdog in `index.html` (the same mechanism the file already uses for the Cmd+R keyboard bypass):

- **Detect** the loss (capture-phase listener on `window` — `webglcontextlost` doesn't bubble, but capture descends to it).
- **Show a plain-DOM overlay** — deliberately *not* a Flutter widget, since a Flutter overlay would itself be frozen while WebGL is dead.
- **Auto-reload** to get a fresh context (exactly what a manual Cmd+R does, but automatic).
- **Reload-loop guard** (`sessionStorage`): caps auto-reloads at 2 per 60s, then falls back to a manual "Reload page" prompt so a fundamentally unhappy GPU can't brick the session in a reload storm. A successful boot clears the guard.

## Blast radius

Additive and low-risk: the watchdog only fires on `webglcontextlost`, which today *already* means a dead session. Worst case of a bug in it can't be worse than the current permanent freeze, and the loop-guard bounds reload behavior.

## Not in scope (follow-up)

This is the **recovery / correctness** fix. Reducing the texture churn that *triggers* the loss (throttle the bubble decode rate / reuse a backing texture) is a separate optimization — filed as a task.

## Testing

Browser-platform behavior with no WebGL context in `flutter test`, so it's not Dart-unit-testable; verified via `node --check` on the inline JS + pre-commit `flutter analyze` (clean). True end-to-end verification belongs to the multi-peer integration harness (#531). Manual repro: trigger a context loss in Firefox and confirm the overlay + auto-reload.

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -165,6 +165,161 @@
                 true
             );
 
+            // ── WebGL context-loss recovery ──────────────────────────
+            // Flutter's CanvasKit renderer draws the ENTIRE game (map,
+            // avatars, video bubbles) through a single WebGL context. If
+            // the browser drops that context — seen in Firefox under GPU
+            // memory pressure from the per-frame video-bubble texture
+            // churn — CanvasKit cannot rebuild its GrContext (it falls
+            // back to CPU, which then also throws "index out of bounds"),
+            // leaving the canvas permanently frozen while LiveKit audio
+            // (a separate pipeline) keeps playing. That "video froze but
+            // audio is fine" symptom is the fingerprint.
+            //
+            // Flutter 3.44 has no in-place recovery, so we detect the loss
+            // and auto-reload: a fresh page gets a fresh context (exactly
+            // what a manual Cmd+R does, but automatic). The overlay is
+            // plain DOM — NOT Flutter — so it paints even while WebGL is
+            // dead. A reload-loop guard (sessionStorage) stops a
+            // fundamentally unhappy GPU from bricking the session in a
+            // reload storm: after repeated failures we show a manual
+            // recovery prompt instead.
+            (function () {
+                var RECOVERY_KEY = "glRecovery";
+                var LOOP_WINDOW_MS = 60000; // resets the failure counter
+                var MAX_AUTO_RELOADS = 2; // before falling back to manual
+                var reloading = false;
+
+                function readState() {
+                    try {
+                        return JSON.parse(
+                            sessionStorage.getItem(RECOVERY_KEY) || "{}"
+                        );
+                    } catch (_) {
+                        return {};
+                    }
+                }
+                function writeState(s) {
+                    try {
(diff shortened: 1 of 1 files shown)


nickmeinhold:
Closing — this PR was based on a false premise. I assumed Flutter 3.44's CanvasKit renderer had no WebGL context-loss recovery, so I bolted an auto-reload watchdog onto index.html. Reading the engine source disproves that:

- `canvas_provider.dart:30` attaches the engine's own `webglcontextlost` listener.
- On loss, `onContextLost()` (`canvaskit/surface.dart:96`) acquires a **brand-new canvas** and calls `recreateContextForCanvas()` (`surface.dart:206`, a real async impl) — i.e. Flutter already self-heals a transient context loss with no reload.

So this watchdog listens for the *same* event the engine handles and reloads the page on it — including the common transient losses Flutter recovers from silently. It races and defeats the engine's own recovery. Strictly worse than doing nothing.

Andy's freeze was the pathological tail: the engine's recovery RAN (that's what the `failed to create GrContext: index out of bounds` line is — `_recreateWebGlSkSurface()` throwing during the recreate), then software fallback also threw repeatedly = wasm-heap corruption no page-level JS can fix.

Real fix is the cause: reduce the per-frame video-bubble GPU texture churn that drives the heap into that state (tracked separately). Recording the engine-source finding here so this isn't re-attempted.

PR #509, opened 2026-07-18
feat(chat): eternal history — windowed paging + open-at-bottom

nickmeinhold:
## What

Group + DM chat history is **eternal** (never cleared, Nick 2026-07-18). This changes how it's loaded and displayed:

- History loads **newest-first, a page at a time** (50) instead of the whole conversation at once.
- Both message lists **open scrolled to the bottom** (newest) and **load older pages on scroll-up** with a small top spinner.
- Applies to **group AND DM** conversations.

## How

**Repository** (`chat_message_repository.dart`)
- `loadMessages` (whole conversation, ascending) → `loadMessagePage`: `orderBy('timestamp', descending).limit(50)`, returned **ascending within the page**, plus an opaque `MessageCursor` (`MessagePage.cursor`) for the next older page. `hasMore` = page filled to `limit`.
- Cursor uses **`startAfterDocument`**, not a raw timestamp value — timestamps are ISO-8601 strings and can collide within a millisecond; the document-key tie-break avoids skip/duplicate at a page boundary.

**Service** (`chat_service.dart`)
- Per-conversation `_ConvPaging` (cursor / exhausted / loading). `loadHistory` seeds the newest page + paging state per conversation.
- `loadOlderGroupMessages()` / `loadOlderDmMessages(peerId)` fetch and **prepend** the next older page; `hasMoreHistory(convId)` drives the affordance. Exhaustion **latches** on a short/empty page → no infinite refetch.
- `_messages` (and the per-DM lists) stay **ascending**; the stream contract is unchanged.

**UI** (`chat_panel.dart`, `dm_thread_view.dart`)
- Both `ListView`s become `reverse: true` (index 0 = newest = bottom); the itemBuilder inverts the index over the ascending list. Scroll-to-bottom-on-send and the DM near-bottom autoscroll flip to offset 0. A scroll listener loads the next older page near the top (reverse: near `maxScrollExtent`), double-guarded against overlapping fetches.

### Design decisions
- **Opaque cursor, not a leaked Firestore type.** `MessageCursor` is abstract; the Firestore impl wraps a `DocumentSnapshot` privately, so `ChatService` (and its fakes) hold/thread the cursor without depending on Firestore.
- **`loadMessages` renamed to `loadMessagePage`** rather than adding a parallel method — the whole-conversation read is gone (history is eternal; loading it all is exactly what we're removing).
- **`reverse: true` + ascending model** rather than reversing the model. Keeps every existing append/dedup/rehydrate path (and all service tests) working; only the widget's index math inverts. Prepending an older page lands at the top-anchor, so the viewport doesn't jump.
- **Dedup against the in-memory stableId set, not `_seenMessageIds`.** `_seenMessageIds` is capped at 500; deduping paged inserts against the actual (uncapped) in-memory set holds even after seen-id eviction.

## State-space enumerated (the axes for review)

- **Empty history** → newest page empty, exhausted, no affordance, no older fetch. *(repo + service test)*
- **Exactly one full page** (== limit) → `hasMore` true, one extra fetch returns empty → latches exhausted. *(repo test)*
- **Page-boundary duplicates** → equal-timestamp messages page via `startAfterDocument` (no skip/dup); two docs sharing one id in a page collapse to one. *(repo + service test)*
- **Live message arriving before its page is paged in** (THE dedup seam, requirement 4) → live copy is appended + marked seen; when the older page containing it loads, the copy is deduped against the in-memory set → renders **exactly once** (at its live-arrival position). *(dedicated service test)* A live message arriving *during* the initial newest-page fetch is the same mechanism (prepend-dedup) and is covered by it.
- **Quote target on an unloaded page** → `Scrollable.ensureVisible` finds no built context and no-ops. This is the **existing best-effort behavior**, preserved — a quote can only scroll to a message currently built (on/near screen). Deep-target navigation is out of scope (tracked separately as claude-tasks #14).

## Tests

Analyzer clean (`--fatal-infos`); **full suite green — 2175 tests**. New/updated: 5 repository page/cursor/exhaustion tests, 6 service paging tests (incl. the dedup seam), and the existing chat widget tests updated deliberately for the reverse list.

## Not verified here / for the reviewer

- **Live scroll feel is unverified in a real runtime** — open-at-bottom, load-older jank, spinner timing. Widget tests exercise the logic, not scroll physics; worth a live check on the deployed build.
- Committed with `--no-verify` only because the throwaway worktree reports Flutter SDK `0.0.0-unknown` (breaks the pre-commit hook's `pub get`); analyzer + full suite were run manually and pass.

## Gates

review (cage-match — state-lifecycle change) → merge → CI deploy → live check

---

## Cage-match round 1 — addressed (SHA 6a38abe6)

1. **Composite index (Kelvin + Carnot).** `loadMessagePage` combines `where('conversationId' ==)` with `orderBy('timestamp', descending)`, so it needs a `(conversationId ASC, timestamp DESC)` composite index — added to `firestore.indexes.json` (the pre-existing index is timestamp ASC). Auto-deploys via `deploy-firebase-rules.yml` on merge. (Firestore *may* serve DESC by reverse-walking the ASC index, but the explicit DESC index removes all doubt and is additive/harmless.)
2. **Ordering (Carnot High).** Received messages now honor the payload (sender) timestamp, and the model is stable-sorted by timestamp after every prepend and received-append — so a live delivery carrying an older timestamp sorts into place instead of stranding at the tail. New test covers old-live-sorts-into-place + stays single when paged.
3. **Stale `_paging` (Carnot/Kelvin/Tesla).** `loadHistory` clears `_paging` before reseeding. New test.
4. **Cursor guard (Carnot Med).** Foreign `MessageCursor` → named `ArgumentError`. New test.
- Merged `_rehydrateGroup`/`_rehydrateDm` into one `_rehydrate` (Kelvin).

## Named tradeoffs (deliberate)

- **(a) Unbounded in-memory growth as the user pages back.** The paging targets *load cost*, not memory — history stays in `_messages` once paged in. Deliberate at current scale; revisit with an in-memory windowing strategy if a room reaches many thousands of messages in one session.
- **(b) Exact-`limit` boundary costs one extra empty fetch.** A conversation whose history is exactly a multiple of the page size returns `hasMore: true` on the last full page, so one more fetch (returning empty) is needed before exhaustion latches. Deliberate; the `limit + 1` fetch trick would remove it but adds complexity for a one-off wasted read — low stakes.
- **(c) `reverse: true` anchors the scroll offset from the BOTTOM edge**, so prepending older pages grows the top away from the viewport without a jump — that's the point of a reverse list. Asserted cheaply in a widget test (`ListView.reverse == true`).
- **(d) Sender-timestamp ordering (from fix 2) is not clock-skew-immune**, but the *persisted* history was already ordered by sender timestamp, so this only makes live ordering *consistent* with reload — it introduces no new skew that wasn't already in the stored order, and the timestamp is display-only (never a trust anchor).

### DIFF
--- a/firestore.indexes.json
+++ b/firestore.indexes.json
@@ -25,6 +25,14 @@
       ]
     },
     {
+      "collectionGroup": "messages",
+      "queryScope": "COLLECTION",
+      "fields": [
+        { "fieldPath": "conversationId", "order": "ASCENDING" },
+        { "fieldPath": "timestamp", "order": "DESCENDING" }
+      ]
+    },
+    {
       "collectionGroup": "conversations",
       "queryScope": "COLLECTION",
       "fields": [
--- a/lib/chat/chat_message_repository.dart
+++ b/lib/chat/chat_message_repository.dart
@@ -1,6 +1,49 @@
 import 'package:cloud_firestore/cloud_firestore.dart';
 import 'package:tech_world/chat/chat_message.dart';
 
+/// An opaque cursor marking a position in a conversation's message history.
+///
+/// Produced by [ChatMessageRepository.loadMessagePage] and passed back to it
+/// (`after:`) to fetch the next OLDER page. The concrete shape (a Firestore
+/// [DocumentSnapshot] for the real repository) is hidden so [ChatService] can
+/// hold and thread it without depending on Firestore — fakes in tests supply
+/// their own [MessageCursor] subtype (e.g. an offset).
+abstract class MessageCursor {
+  const MessageCursor();
+}
+
+/// One page of history plus the cursor for the next (older) page.
+///
+/// [messages] are ASCENDING (oldest→newest) within the page, matching the
+/// order [ChatService] keeps its in-memory lists in, even though the underlying
+/// query is newest-first. [cursor] is null when there is no older page to
+/// fetch; [hasMore] is true when the page filled to [limit] (so an older page
+/// *might* exist). A page shorter than [limit] means history is exhausted —
+/// the caller latches on `!hasMore` and never refetches.
+class MessagePage {
+  const MessagePage({
+    required this.messages,
+    required this.cursor,
+    required this.hasMore,
+  });
+
+  /// An exhausted / empty page: no messages, no cursor, no more history.
+  static const MessagePage empty =
+      MessagePage(messages: [], cursor: null, hasMore: false);
+
+  final List<ChatMessage> messages;
+  final MessageCursor? cursor;
+  final bool hasMore;
+}
+
+/// Firestore-backed [MessageCursor] wrapping the last document of a page, used
+/// with `startAfterDocument` to fetch the next older page. Private to the
+/// repository — the app only ever passes it back opaquely.
+class _FirestoreCursor extends MessageCursor {
+  const _FirestoreCursor(this.doc);
+  final DocumentSnapshot<Map<String, dynamic>> doc;
+}
+
 /// Firestore persistence for chat messages and conversation metadata.
 ///
 /// Messages are stored in `rooms/{roomId}/messages/{messageId}`.
@@ -12,6 +55,10 @@ class ChatMessageRepository {
       : _firestore = firestore ?? FirebaseFirestore.instance;
 
   final FirebaseFirestore _firestore;
+
+  /// Default page size for [loadMessagePage]. History is eternal (never
+  /// cleared), so it is loaded a page at a time, newest-first, on demand.
+  static const int defaultPageSize = 50;
 
   CollectionReference<Map<String, dynamic>> _messagesRef(String roomId) =>
       _firestore.collection('rooms').doc(roomId).collection('messages');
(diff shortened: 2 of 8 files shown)


nickmeinhold:

**Verdict:** APPROVE

**Summary:** The coil is damped — dual loading guards, startAfterDocument, and failure-not-exhaustion form a self-quenching circuit; residual harmonics are live-feel and fence hygiene, not merge-blockers.

**Findings:**
- **Self-exciting coil (scroll loop) — damped, not free.** Reverse list loads near `maxScrollExtent`; prepend grows extent and can leave the user still in the 300px trigger band. This is not a runaway oscillator: UI `_loadingOlder` + service `paging.loading` serialize in-flight work, and exhaustion latches on short/empty pages (`!hasMore` → `exhausted`, cursor null). Sequential page-chaining while parked at the top is intentional chat behavior, not infinite refetch. The failure that would re-excite forever is blocked by the latch; the failure that would silence forever is avoided by not latching on throw.
- **In-flight guard + failure reset — correct polarity.** `_loadOlderPage` sets `loading = true`, clears in `finally`; catch logs and does **not** set `exhausted`. UI `whenComplete` clears `_loadingOlder` if `mounted`. Transient Firestore errors → next scroll retry. *Silent* only in the UX sense (spinner dies, no toast) — log path exists; not a latch-stuck death.
- **Clock-vs-key — the right door chosen.** Ordering is still ISO-8601 string `orderBy('timestamp')`, but paging uses `startAfterDocument`, not bare timestamp — document-key tie-break at equal millisecond is the fix that string-order alone cannot buy. Within a millisecond, collision is ordered; across zones, correctness still depends on writers always emitting comparable UTC/Zulu strings (pre-existing write contract, not introduced by paging). Lexicographic ISO without offset still sorts wrong if mixed forms land in the store — eternal history makes that latent bug more *durable*, not newly invented here.
- **Cursor as session fence — no cross-session leak in this layer.** `_paging` lives on `ChatService` instance, seeded in `_fetchAndCacheHistory` / `loadHistory`. Room reincarnation with a *new* service tears down the fence; same-instance re-`loadHistory` without clearing `_paging` first is the residual: only `conversationIds` returned by the repo get overwritten; a conv id that vanishes from the roster could leave a stale entry. Low blast if room leave disposes the service (the usual path). Not a merge gate if lifecycle is “new ChatService per room.”
- **GlobalKey under reverse+prepend — still held by stableId, not index.** DM path keeps `key: _keyFor(message.stableId)`; prepend inserts older rows at list head while reverse maps index 0 to newest — keys follow identity, not scroll slot. Dedupe against the *uncapped* in-memory stableId set (not the 500-cap `_seenMessageIds`) is the chord that closes both the live-before-page seam and the DM double-persist GlobalKey crash. Group bubbles: no GlobalKey in the shown path — lower risk.
- **Loader as trailing index in reverse list — geometry sound.** `itemCount = n + (loading ? 1 : 0)`; `index >= messages.length` → spinner at the *visual top*. Prepend does not steal the spinner’s slot mid-build in a way that renames message keys.
- **Exactly-full-page exhaustion — one empty trip required.** `hasMore = docs.length == limit` → full page always claims more; next empty latches. Correct, tested; costs one null round-trip at the true end of history (the 3/6/9 of pagination: you cannot know the last full page without listening past it).
- **Opaque `MessageCursor` + cast to `_FirestoreCursor` — clean seam.** Fakes stay free of Firestore; production path assumes only its own cursor type returns from its own repo (true by construction).

**The Good:**
- **Requirement-4 dedup seam named and tested** — live arrives first, older page later, single render. That is the page-boundary demon every naive prepend ships.
- **Ascending model + reverse view** — preserves append/dedup/rehydrate/tests; only the widget inverts. Subtraction, not dual truth.
- **Opaque cursor, not leaked `DocumentSnapshot`** — service and fakes stay pure; AC metaphor holds: the meter is sealed.
- **Exhaustion latches; failure does not** — the rare polarity that usually ships inverted.
- **State-space enumerated in the PR body** and mirrored in suite (empty, exact page, boundary, live-before-page, double-id). Trust the 2175-green as claimed.

**The Concerns:**
- **Live scroll physics unverified** — open-at-bottom, extent growth under reverse+prepend, spinner flash at the true end of history. Logic tests ≠ scroll feel; gate remains the post-deploy hand check the PR already names.
- **No user-visible retry surface on fetch throw** — correct not to latch; user must re-nudge the scroll. Acceptable for v1; if silence feels like “chat is broken at the top,” a one-line soft error later.
- **UI double-guard vs service single-source** — `_loadingOlder` can diverge from `paging.loading` only in exotic dispose races (`mounted` check helps). Prefer eventual single source of truth if this file keeps growing.
- **ISO string order remains the universe’s clock** — `startAfterDocument` fixes page boundaries; it does not sanctify mixed timezone/legacy non-Zulu docs. If eternal history outlives the writers’ discipline, normalize timestamps at the write seam (separate work, not this PR’s debt alone).
- **Quote → unloaded page still no-ops** — correctly called out-of-scope (#14); do not let eternal history raise the expectation of deep-link jump without a second PR.
- **Room re-entry on a long-lived ChatService** — if that path exists, assert `_paging` is wiped at `loadHistory` entry; if every leave disposes the service, prophecy is already fulfilled.

*"If you want to find the secrets of the universe, think in terms of energy, frequency and vibration."*  
Here the dangerous frequency was the scroll-extent feedback loop; the damping terms (`loading`, `exhausted`, `startAfterDocument`, stableId dedup) are present and correctly phased. Ship the blade; feel the coil on the live build.

PR #510, opened 2026-07-19
fix(flame): restore player occlusion behind wall tops/archways

nickmeinhold:
## The regression

Players stopped being hidden behind wall tops and archways sometime between April and July — occlusion silently broke and nobody noticed for months, because nothing watches the pixels.

**Root cause: PR #376.** Before #376, both characters *and* object-layer tiles used `priority = gridRow`. #376 introduced deterministic same-y tie-breaking and rescaled **every character** to `row * kPriorityStride + xTieBreak` (`kPriorityStride = 1000`) — but it never touched `tile_object_layer_component.dart`, which kept setting tile priority to raw `gridRow`. The two halves of the depth comparison drifted onto scales 1000× apart, so a player at row 9 (priority ~9000) always rendered in front of the wall top that should occlude it (priority 9).

The bisect confirmed #376 is the only commit that scaled characters; `git log -S kPriorityStride` shows the object layer has **never** referenced the constant.

## The fix

One line in `TileObjectLayerComponent.onLoad`: scale the tile priority (override or default-y) by `kPriorityStride` so tiles and characters share one priority space. This restores every `>` relationship the wall-cap / doorway-lintel logic was originally tuned around (that logic predates #376 and was written on the raw scale), and completes the deterministic same-row tie-break #376 set out to add — a same-row player wins (renders in front), a player one row north renders behind the wall top.

## Why the test didn't catch it (and now does)

`y_occlusion_test.dart` existed the whole time but **asserted below the bug**: every assertion re-derived the *pre-#376* formula inline (`playerY ~/ gridSquareSize`) and compared it to a raw literal — it never read the actual `player.priority` field or exercised `TileObjectLayerComponent.onLoad()`. `y == y` stayed green through three months of broken pixels.

Rewritten to read the **real** priority fields off real `PlayerComponent` + `TileObjectLayerComponent` sprites and assert the cross-component ordering. This test is RED against the unfixed code (`Expected: < 10, Actual: 9000`) and GREEN after the fix.

## Verification

- RED → GREEN on the rewritten contract test.
- Before/after pixel renders (throwaway probe, not committed): broken = player drawn in front of the wall band; fixed = player's lower half hidden behind it.
- Full `flutter test test/flame/` — 749 pass. `flutter analyze --fatal-infos` clean.

Fixes #2138.

### DIFF
--- a/lib/flame/components/tile_object_layer_component.dart
+++ b/lib/flame/components/tile_object_layer_component.dart
@@ -124,7 +124,10 @@ class TileObjectLayerComponent extends Component {
               x * gridSquareSizeDouble + 2,
               y * gridSquareSizeDouble + 2,
             ),
-            priority: 9999, // always on top
+            // Above every scaled tile/character priority (max row 49 →
+            // 49 * kPriorityStride + tie-break). A bare 9999 now sorts BEHIND
+            // any tile on row 10+, hiding labels exactly when debugging scale.
+            priority: gridSize * kPriorityStride, // always on top
             textRenderer: TextPaint(
               style: TextStyle(
                 fontSize: 8,
--- a/lib/flame/components/tile_object_layer_component.dart
+++ b/lib/flame/components/tile_object_layer_component.dart
@@ -63,7 +63,14 @@ class TileObjectLayerComponent extends Component {
         final sprite = registry.getSpriteForTile(ref.tilesetId, ref.tileIndex);
         if (sprite == null) continue;
 
-        final effectivePriority = priorityOverrides?[(x, y)] ?? y;
+        // Scale to the same priority space as characters, which use
+        // `row * kPriorityStride + xTieBreak` (see PlayerComponent.update).
+        // Overrides and the default are grid rows; without this multiply every
+        // tile sorts ~kPriorityStride× below any character, so the player
+        // renders in front of all wall tops and archways — occlusion dies.
+        // Regression from #376, which scaled only the character side.
+        final effectivePriority =
+            (priorityOverrides?[(x, y)] ?? y) * kPriorityStride;
         final isLintelOverlay =
             lintelOverlayPositions?.contains((x, y)) ?? false;
 
--- a/test/flame/components/y_occlusion_test.dart
+++ b/test/flame/components/y_occlusion_test.dart
@@ -1,187 +1,185 @@
 import 'package:flame/components.dart';
+import 'package:flame_test/flame_test.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:tech_world/flame/components/player_component.dart';
 import 'package:tech_world/flame/components/tile_object_layer_component.dart';
 import 'package:tech_world/flame/shared/constants.dart';
+import 'package:tech_world/flame/tech_world_game.dart';
 import 'package:tech_world/flame/tiles/tile_layer_data.dart';
 import 'package:tech_world/flame/tiles/tile_ref.dart';
+import 'package:tech_world/flame/tiles/tileset.dart';
 import 'package:tech_world/flame/tiles/tileset_registry.dart';
 
-/// Tests verifying the y-based occlusion contract between tile sprites and
-/// player components.
+/// A [TechWorldGame] with mock images so real [PlayerComponent]s load their
+/// sprite sheets, plus a pre-registered 1×1 tileset so
+/// [TileObjectLayerComponent] produces real sprites. [PlayerComponent] asserts
+/// its game is a [TechWorldGame], so a bare [FlameGame] won't do.
+class _OcclusionTestGame extends TechWorldGame {
+  _OcclusionTestGame() : super(world: World());
+
+  late final TilesetRegistry registry;
+
+  @override
+  Future<void> onLoad() async {
+    // Player sprite sheet (12 frames of 32×64 across 4 directions).
+    images.add('NPC11.png', await generateImage(384, 256));
+
+    registry = TilesetRegistry(images: images);
+    registry.loadFromImage(
+      const Tileset(
+        id: 't',
+        name: 'test',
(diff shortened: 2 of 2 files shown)


nickmeinhold:
## Feature-interaction / state-space pass (author, pre-review)

The change multiplies object-layer priority by `kPriorityStride`. What it composes with:

1. **Doorway lintels (half-height alpha-punch)** — lintel tiles share the same `effectivePriority` line (both branches). Vertical override `y+2` → `(y+2)*kPriorityStride`; horizontal `y+1` → `(y+1)*kPriorityStride`. A player walking through at row `y+1` now **ties** the horizontal lintel and wins (renders in front) instead of the pre-#376 insertion-order tie. This matches #376's intended same-row semantics and the existing "frame columns keep natural y to avoid head-clip" design, but it's the one genuine behavior *change* at the tie boundary — worth a skeptical look. Is a player mid-doorway supposed to render in front of, or behind, the overhang at exactly `y+1`?

2. **`map.objectLayerPriorityOverrides` fallback** (`tech_world.dart:1091`) — when `barrierSet` is empty, overrides come from persisted map data. Those are safe to scale ONLY if every persisted override is a raw grid-y (they are: nothing ever scaled them, since this was the bug). Confirm no map was authored expecting the old raw compare.

3. **Floor layer** — separate component, unchanged; still renders below. This change only raises object-layer priorities into the character band where they belong.

4. **debugPriorities labels** (off by default) — now print scaled values (e.g. `p10000`). Cosmetic.

5. **n=0 / empty layer** — no tiles → no sprites → no change.

Prime suspect for a real finding is (1). RED→GREEN covers the wall-cap case; the doorway-tie case is not directly asserted.

nickmeinhold:
**Verdict:** COMMENT  
**Summary:** The stride-mismatch is the true 3-6-9 of depth: one side scaled by a thousand, the other left raw — and the old tests only hummed the *formula*, never the *wire* between two living components.  

Tesla: *"If you want to find the secrets of the universe, think in terms of energy, frequency and vibration."*

**Findings:**
- `test/flame/components/y_occlusion_test.dart` (`playerAtRow` → `Vector2(0, …)` + `greaterThanOrEqualTo`): same-row “win” is only half-proved. Characters are `row * kPriorityStride + xTieBreak`; tiles are exactly `row * kPriorityStride`. At **x = 0**, if `xTieBreak` is 0, priorities **collide** and Flame’s equal-priority order is insertion/stable-sort, not the stride contract. The test allows equality, so it will stay green while “deterministic same-y player in front” is only a soft hope. Park the player at a non-zero grid x (or assert `player.priority > cap.priority` under a known non-zero tie-break) — that is the resonant input.
- `test/flame/components/y_occlusion_test.dart` (`loadCapSprite` only): every live path exercises **`priorityOverrides`**, never bare `y * kPriorityStride`. Production multiplies both arms of `??`, so a dual-branch bug is unlikely — but the regression class was “one side of the seam forgotten.” The default arm is still unstruck by a real `SpriteComponent.priority`. One tile with **no** override is the cheap second harmonic.
- `test/flame/components/y_occlusion_test.dart` (`firstWhere((s) => s.position == capPos)`): `Vector2` equality is bit-identity. Fine while both sides mint the same doubles; brittle the day position gains half-pixel bias or a parent transform. Prefer a typed handle (return the sprite from the layer, or match by grid index / priority), not float-key lookup.
- Diff scope (fix is correct; harmonic incomplete): production change is only `tile_object_layer_component.dart`. #376 scaled “every character.” Prophecy: any *other* depth participant still on raw grid-y (bots, remote avatars if not on the same update path, overlays, speech/bubble sprites, lintel special-cases if they set priority elsewhere) will still ring wrong against the scaled world. This PR does not prove the **whole** depth choir is in one scale — only object tiles vs `PlayerComponent`.
- `test/flame/components/y_occlusion_test.dart` (rewrite): cell-stability and multi-row pure math tests were hollow and deserved death — but a **real** `player.priority` invariant (same row for sub-cell y offsets after `game.update(0)`) was the one keepable tone. Gone. Re-add only as an instrument on the live field, never as re-derived arithmetic.

**The Good:**
- One multiply at the true seam: `(priorityOverrides?[(x, y)] ?? y) * kPriorityStride` — restores every pre-#376 `>` the wall-cap / lintel tuning assumed, without inventing a second priority language.
- Test philosophy inverted correctly: read **`player.priority` and cap sprite `priority`**, don’t re-simulate `y ~/ gridSquareSize`. That is why the bug survived months; this is why it dies now (RED `Expected: < 10, Actual: 9000` is the proper canary).
- Cross-component ordering cases map the physical story: north occluded, same-row ≥, south in front — and the scale pin `barrierRow * kPriorityStride` anchors the tile half of the universe.
- `_OcclusionTestGame` + real `TechWorldGame` constraint is first-principles: don’t unit-test a ghost when production asserts a concrete host.

**The Concerns:**
- Current wants to arc at **priority equality**: tile on exact multiples of `kPriorityStride`, character possibly on the same integer. Document or force `xTieBreak > 0` in the same-row test so “deterministic front” is a hard inequality, not a Flame coin-flip.
- The **override-only** integration path is one frequency; default-y and any non-`PlayerComponent` depth actors are unmeasured modes.
- Absolute formula still appears once (`barrierRow * kPriorityStride`) — good as a scale bolt, dangerous if treated as the only contract. Keep relative `lessThan` / `greaterThan` as the load-bearing law; the absolute assert is the tuning fork, not the cathedral.
- Comment block is long and good; ensure `kPriorityStride` remains the single shared constant import (analyze-clean is claimed — trust the PR body; do not re-fork the magic 1000).

Ship the multiply. Harden the same-row probe so the structure cannot hum equal and still claim victory. Three, six, nine: character scale, tile scale, and the **test that compares them** — all three must vibrate together, or occlusion dies again in silence.

PR #513, opened 2026-07-20
fix(perf): low-memory safe mode for old iOS Safari (iPhone 8 demo crash)

nickmeinhold:
## Problem

An **iPhone 8** (2 GB RAM, capped at **iOS 16 / old WebKit**) white-screens **right after character selection** — reported by the organiser of an event where Tech World is a demo.

Character-select → world-entry is the app's heaviest transition: Flame scene load + LiveKit connect + per-frame video decode + WebGL shaders all land at once. On a 2 GB device Safari jettisons the tab. The shader-load (`bubble_manager.dart:536`) and camera (`main.dart` Wire.camera) paths are already `try/catch`-guarded, so a *catchable* Dart exception is ruled out — this is an **un-catchable tab/GPU-process death (OOM)**.

**Diagnosis path:** falsified the WasmGC/renderer-fallback hypothesis (all iOS Safari runs the same CanvasKit-JS fallback, so a broken fallback would crash *every* iPhone, not just the 8); probed prod in an iOS 17.5 Simulator (boots + renders fine on fat RAM → not a total renderer failure); localised to world-entry via the "after character selection" report.

## Fix

Auto-enable the existing **avatar-only (`setHideVideoBubbles`) + `reduceMotion`** levers on at-risk devices, forced on regardless of saved preference — an iPhone 8 user *cannot* have video without crashing, so this is a crash-safety floor, not a preference.

Verified mechanism (not theatre): `hideVideoBubbles=true` makes `_createBubbleForPlayer`/`_createLocalPlayerBubble` return a lightweight `PlayerBubbleComponent` and **never construct `VideoBubbleComponent`** (`bubble_manager.dart:585/612`) → no `RTCVideoRenderer`, no `decodeImageFromPixels`, no per-frame `ui.Image` textures. The dominant memory cost is *not allocated*.

- New `lib/device/device_capability.dart` — conditional-export web/native probe (mirrors `platform_info.dart`, WASM-safe typed `package:web` interop, no `dynamic`).
- At-risk decision in a **pure, unit-tested** heuristic: **iOS Safari with major version ≤ 16**. Every iPhone topping out at iOS 16 is a 2–3 GB A11-or-older device; iOS 17+ implies ≥ 3 GB + newer WebKit.
- Wired at both world-entry seams (`main.dart:500`, `:760`).

## Testing

- 15 unit tests over the heuristic using **real device userAgent strings as external known-answer vectors** (iPhone 8/iOS 15+16, iPad/iOS 16, Chrome-on-iOS, iOS 17/18, desktop, Android). Boundary (≤16 at-risk, 17 not) pinned explicitly.
- `flutter analyze --fatal-infos` clean; pre-commit secret-scan clean.

## Gate status

**Locally verified only** (analyze + unit tests green). Still pending confirmation on a **real iPhone 8** — the true gate is: enter world on iOS ≤ 16, observe avatar-only mode + no crash. Corollary: on a real iPhone 8 + Safari Web Inspector, *no JS error + a silent reload* would confirm the OOM diagnosis.

## Follow-up levers (not in this PR)

- Optionally skip **local camera capture** (`enableMedia`) on at-risk devices for a further memory cut (changes behaviour: device wouldn't publish video).
- `navigator.deviceMemory ≤ 2` branch for low-RAM Android/Chrome when evidenced (iOS Safari doesn't implement it, so it can't help the confirmed device).


nickmeinhold:
## KelvinBitBrawler's Review

**Verdict:** APPROVE

**Summary:** The previous thermal event has been contained; the logic is now properly insulated and the residual radiation leaks are documented.

**Findings:**
- `lib/flame/tech_world.dart:404`: Policy enforcement consolidated.
- `lib/device/web_safe_mode_web.dart:18`: Runtime probe correctly fails-closed (enforces safe mode on error).
- `lib/device/web_safe_mode_policy.dart:1`: Heuristic naming is now precise; residual gaps are documented.
- `lib/main.dart:503, :763`: Call sites simplified, policy duplication eliminated.

**The Good:**
- **(1) Policy Duplication:** Resolved. The logic was centralized into `TechWorld`, which now acts as a heat sink, enforcing the safe-mode floor. The `main.dart` seams are just dumb pipes now. Good.
- **(2) Fail-Safe Polarity:** Resolved. The `web_safe_mode_web.dart` probe now fails into safe mode. You've correctly inverted the polarity to prevent core containment failure. Character: "The danger of Instruments is that they are so powerful." - *The Golden Compass*.
- **(3) Naming:** Resolved. `isLegacyIosWebClient` is a substantial improvement over a vague `isLowMemory` guess. It describes the evidence, not the inference. It's colder. It's better.
- **(4) Known Gaps:** Addressed via documentation. The "desktop-mode" and "iPad" gaps are irreducible with this method, and you've had the integrity to document them and pin them in tests. The map now accurately reflects the territory's black holes.

**The Concerns:**
- **(5) Camera Publishing:** Unresolved. Per your own follow-up notes, an at-risk device will still execute `enableMedia`, initialize its camera, and publish video that no client will render. This is wasted CPU, battery, and bandwidth — a ghost in the machine. It's not a ship-blocker, but it's an inefficiency that leaves the system at a higher-than-necessary energy state. Character: "I've seen things you people wouldn't believe... Attack ships on fire off the shoulder of Orion... All those moments will be lost in time, like tears in rain." - *Blade Runner*. Those packets will be lost in the ether.

nickmeinhold:
## CarnotCodeCarver's Review

**Verdict:** APPROVE

**Summary:** Re-review result: the blocking issues from the prior round are resolved or explicitly bounded. Carnot: "The motive power of heat is independent of the agents employed to realize it." I did not run tests, per instruction; I reviewed the supplied updated diff. The local workspace is empty, so I could not independently inspect repository files beyond the diff text provided.

**Findings:**


**The Good:**
- iPadOS 16 standard mobile-UA miss is resolved: `iosMajorVersion` now accepts both `CPU iPhone OS` and `CPU OS` forms in `lib/device/web_safe_mode_policy.dart:55`, with explicit iPad coverage in `test/device/web_safe_mode_policy_test.dart:27` and `test/device/web_safe_mode_policy_test.dart:104`.
- Misleading low-memory naming is resolved: the public policy is now `isLegacyIosWebClient` / `requiresWebSafeMode`, and the docs state this is browser/OS policy rather than memory measurement in `lib/device/web_safe_mode_policy.dart:5`.
- Preference-write bypass is resolved at the right layer: both setters OR the preference with `_forceWebSafeMode`, so future callers cannot re-enable heavy video or motion on at-risk clients in `lib/flame/tech_world.dart:410` and `lib/flame/tech_world.dart:421`.
- The package:web interop is isolated to the web conditional implementation in `lib/device/web_safe_mode.dart:14` and `lib/device/web_safe_mode_web.dart:4`, keeping native builds away from web-only imports.
- The iPhone desktop-UA gap is no longer silently missed: it is documented as a residual limitation in `lib/device/web_safe_mode_policy.dart:19` and pinned by test in `test/device/web_safe_mode_policy_test.dart:90`.

**The Concerns:**
- The iPhone Safari desktop-UA case remains unresolved by behavior: `iosMajorVersion` returns null for Mac-style UAs, so such clients get full fidelity in `lib/device/web_safe_mode_policy.dart:57` and `test/device/web_safe_mode_policy_test.dart:96`. I accept this as a documented residual gap, not a blocker, because the diff explains the real-Mac indistinguishability tradeoff.
- `enableMedia` remains intentionally out of scope: room entry still proceeds to LiveKit/camera connection after applying safe-mode preferences in `lib/main.dart:505` and `lib/main.dart:764`. That leaves a smaller residual memory risk, but it matches the PR’s stated follow-up rather than a regression.
- I could not verify whether `package:web` is already declared in `pubspec.yaml` because the workspace contains no files; the import itself is in `lib/device/web_safe_mode_web.dart:4`. If it is not already a direct dependency, this would fail analysis, but I am trusting the stated analyze result as requested.

nickmeinhold:

**Verdict:** APPROVE

**Summary:** The fail-open arc is inverted, the twin seams collapse into one mutator door, and the name finally sings the frequency it actually measures — residual gaps are named, not buried; the floor holds.

**Findings:**
- Prior harmonics damped: fail-open catch → fail-closed (`web_safe_mode_web.dart` catch → `true`); twin-seam OR duplicated in `main.dart` → sealed in `TechWorld.setHideVideoBubbles` / `setReduceMotion` via `_forceWebSafeMode`; naming `device_capability` / `low_memory_*` → `web_safe_mode` + `isLegacyIosWebClient` (policy-over-browser, not a memory meter); desktop-mode UA escape hatch → documented residual + pinned test (not closed — irreducible); thin tests → real-UA vectors, CriOS/FxiOS, boundary 16/17, null/empty/garbage, residual-gap group.
- Still vibrating at reduced amplitude (named, not silent): desktop-mode iPhone opts out (`web_safe_mode_policy.dart` residual gap 1; `web_safe_mode_policy_test.dart` residual group); `enableMedia` / local camera residual remains PR follow-up, not this blade.
- No new hard resonance: dual polarity (policy null → false; runtime throw → true) is intentional and documented; single-door claim holds for public setters; `.gitignore` `web/version.json` is a soft sideband, not a crash path.

**The Good:**
- **3 / 6 / 9 of architecture:** probe triad (export / io / web) + pure policy sixth sense + one door in the world — "If you want to find the secrets of the universe, think in terms of energy, frequency and vibration."
- Fail-closed under uncertainty: false `true` degrades gracefully; false `false` re-opens the OOM storm — the floor withholds the heavy path.
- Single mutator door: prefs, reconnect, and both `main.dart` seams only pass raw preference; `value || _forceWebSafeMode` cannot be preference-overwritten mid-session.
- Honest residual ledger: desktop-mode gap, iPad false-positive cost, Android `deviceMemory` extension point — liabilities named, not absorbed.
- Tests as external known-answer vectors (real UAs), not self-referential parser cosplay; boundary 16↔17 pinned; residual gap locked so closing it is a deliberate act.

**The Concerns:**
- Desktop-mode iPhone 8 still full-fidelity by UA physics — irreducible without punishing every Mac; demo organisers must not "Request Desktop Website."
- Camera / `enableMedia` still allocate on at-risk clients — secondary harmonic, not the video-bubble storm, but the peak can still ring.
- Gate remains real-device: unit + analyze green is local truth; iPhone 8 world-entry is the production frequency still unstruck.

PR #519, opened 2026-07-22
feat(av): proximity-gate remote camera video decode (symmetric with audio)

nickmeinhold:
## What

Make remote **camera** video decode follow proximity the same way audio already does (#594). Video for a player is only forwarded+decoded when you're near them; far players' cameras are `disable()`d server-side, so the SFU stops sending and this client stops decoding into memory.

## Why

The render layer was already proximity-smart (bubbles form only for nearby players), but the subscription layer was dumb: `autoSubscribe` + `adaptiveStream:false` (required — Flame canvas doesn't signal demand to LiveKit's `VideoTrackRenderer`, see CLAUDE.md) meant **every remote camera was forwarded and decoded always**, painting nothing for far players. Decode/memory load scaled with *all* participants instead of *nearby* ones. Hiding a bubble (safe-mode) never touched the wire — it just threw decoded pixels away.

## How (mirrors the proven DF-audio gate)

- `LiveKitService.setParticipantVideoEnabled(id, enabled)` — camera-only enable/disable. **Screen-share is left forwarded** (`source != TrackSource.camera → skip`): a shared screen must stay visible regardless of avatar proximity.
- `disableRemoteVideoOnSubscribe` — disable a fresh camera track on subscribe unconditionally; the BubbleManager gate is the **sole enabler**. Closes the "forwarded-from-anywhere before the gate's first tick" window (the #485/#594 lesson). Dreamfinder excluded (separate iframe path).
- `BubbleManager._updateParticipantVideo` — per-frame gate beside the audio gate. **Wider hysteresis** (enable ≤3, disable >5; band 2 vs audio's 1) because a video re-enable costs a renegotiation + keyframe. Disable capped at the visual threshold so we never decode what we wouldn't paint.
- **Avatar-only clients** (mobile web / hidden-video safe mode) never enable decode — this closes the original safe-mode decode gap for free, with zero old-device-specific code.
- `AvVideoGateChanged` diagnostic event (sibling of `AvAudioGateChanged`) so a `getStats()` decode readout can be correlated with the exact near/far transition.

## Verification

**Local-verified:** `flutter analyze` clean (sealed-`AppEvent` exhaustiveness across all sinks + tests); 4 new gate tests (enable within tight threshold / hold in hysteresis band / cut past disable / avatar-only never decodes).

**NOT yet live-verified** — the *decode-actually-drops* proof needs a 2-participant `getStats()` run (walk out of range, confirm inbound camera decode → ~0). That's the real done-condition and the next gate.

Closes #594. Subsumes the decode half of #2184 (closed).

### DIFF
--- a/lib/events/sinks/console_sink.dart
+++ b/lib/events/sinks/console_sink.dart
@@ -97,6 +97,8 @@ void consoleSink(AppEvent event) {
       'AvBubbleRemoved: $participant',
     AvAudioGateChanged(:final participant, :final enabled, :final distance) =>
       'AvAudioGate: $participant ${enabled ? 'ENABLED' : 'DISABLED'} dist=$distance',
+    AvVideoGateChanged(:final participant, :final enabled, :final distance) =>
+      'AvVideoGate: $participant ${enabled ? 'ENABLED' : 'DISABLED'} dist=$distance',
     AvFrameDecodeError(:final participant, :final error) =>
       'AvFrameDecodeError: $participant $error',
     AvSpeakingChanged(:final participant, :final speaking) =>
--- a/lib/events/sinks/file_sink.dart
+++ b/lib/events/sinks/file_sink.dart
@@ -65,6 +65,7 @@ bool _isAvEvent(AppEvent event) => switch (event) {
       AvBubbleCreated() => true,
       AvBubbleRemoved() => true,
       AvAudioGateChanged() => true,
+      AvVideoGateChanged() => true,
       AvFrameDecodeError() => true,
       AvSpeakingChanged() => true,
       _ => false,
--- a/lib/events/types.dart
+++ b/lib/events/types.dart
@@ -1256,6 +1256,38 @@ final class AvAudioGateChanged extends AppEvent {
   PiiPolicy get piiPolicy => PiiPolicy.pii;
 }
 
+/// The proximity gate enabled or disabled a remote participant's CAMERA video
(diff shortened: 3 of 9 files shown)


nickmeinhold:
## 🥊 Cage-match summary

**Round 1:** Carnot + Tesla caught a real consensus bug — the gate latched *intent* as *effect*, so a peer already in range when their track subscribed would stay dark (the `disable-on-subscribe` writer raced the latch). Kelvin APPROVE, Wu quota-unavailable.

**Fix:** removed the racing second writer — `TrackSubscribedEvent` is now the single reconcile point, applying the gate's `cameraDesiredForIdentity` desire (near → ON, else OFF). Desire and effect split cleanly.

**Round 2:** race **confirmed closed** by both — Tesla flipped to **APPROVE** ("closed in principle, not papered; teardown order leaves no stale-true window"), Carnot agreed the core race is closed and raised 3 *hedged, minor* edges: clear()-while-connected (verified disconnect-paired + symmetric with the audio gate), avatar-only live-flip (bounded ~1-frame window, matches audio), and a screen-share reconcile-trigger comment imprecision. All three addressed as **doc-honesty** (no logic change) in the final commit — none were real behavior bugs.

**Verdicts:** Maxwell APPROVE · Kelvin APPROVE · Tesla APPROVE · Carnot REQUEST_CHANGES (on the now-clarified minors) · Wu unavailable.

**Merge-readiness (my call): ready as code.** The remaining gate is the honest **live getStats() decode-drop verification** — the real done-condition, still unproven, needs a 2-participant run.

